PR #13, opened 2023-11-15
Initial models and Routes - Agencies/Organizations

as1729:


web/src/components/Agency/AgencyForm/AgencyForm.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <Form<FormAgency>
+      onSubmit={onSubmit}
+      formMethods={formMethods}
+      error={error}
+      className={hasErrors ? 'was-validated' : ''}
+    >
+      {agency && (
+        <div className="row">
+          <Label name="id" className="form-label col-sm-2 col-form-label">
+            Agency Code
```
ed-snodgrass, line 49: Is Agency Code supposed to map to the agency.id? I don't think we need to hold up this PR for this but it is probably worth at least noting this explicitly.

ed-snodgrass (approved):

PR #35, opened 2023-12-07
feat: add User and Role to schema

dysmento:
Expanding our schema with `User` and `Role`.


ed-snodgrass (approved):
Migration ran successfully for me. 

I will say that previously @as1729 was using `created_at DateTime @default(now()) @db.Timestamptz(6)` for timestamps. Not sure if that is important or not, I assume it is fixable later if necessary.

PR #36, opened 2023-12-07
Add User and Role models

dysmento:
This adds the Reporting Periods page, which includes three new models for the schema: InputTemplate, OutputTemplate, and ReportingPeriod.

I generated the SDL I needed for this to work, which included some models that didn't have SDL yet (roles and users).


ed-snodgrass (approved):
Looks good

PR #39, opened 2023-12-08
fix: remove tf-plan conditional in CI workflow

TylerHendrickson:
This PR fixes an issue that is currently blocking certain PRs from running the entire Continuous Integration workflow to completion. The root problem was the presence of an `if:` job conditional that was evaluating to `false`, causing the "Plan Terraform" phase to be skipped. Since generating a successful Terraform plan is a requirement for PRs, they're effectively blocked from merging.

The problematic conditional doesn't seem to be doing anything useful, so it's been removed entirely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,6 @@ jobs:
       - aws-auth
       - build
     uses: ./.github/workflows/terraform-plan.yml
-    if: needs.build.outputs.build-api-image-result == 'success' && needs.build.outputs.build-web-result == 'success' && needs.aws-auth.result == 'success'
     with:
       ref: ${{ github.event.pull_request.head.sha }}
       concurrency-group: run_terraform-staging


.github/workflows/ci.yml
```diff
@@ -78,7 +78,6 @@ jobs:
       - aws-auth
       - build
     uses: ./.github/workflows/terraform-plan.yml
-    if: needs.build.outputs.build-api-image-result == 'success' && needs.build.outputs.build-web-result == 'success' && needs.aws-auth.result == 'success'
```
TylerHendrickson, line 81: Alternate fix:
```yaml
if: needs.build.outputs.build-console-image-result == 'success' && needs.build.outputs.build-web-result == 'success' && needs.aws-auth.result == 'success'
```

As mentioned in the opening comment, I don't believe this conditional is doing anything useful, which is why I removed it altogether. The above line is also suitable, as it gates the conditional on a successful Docker build of the RedwoodJS console image. 

API image builds are currently skipped since they were taking a while to run and aren't actually being deployed anywhere, which is why this check was always evaluating to `false`.

as1729 (approved):

PR #42, opened 2023-12-09
feat: add Subrecipient and Project to the schema

dysmento:
This change should complete the schema. Adds Subrecipient and Project, and their generated SDLs.


ed-snodgrass (approved):
Database tables were generated correctly when I ran the migrate.

Issue #16, opened 2023-11-15
User Management: Editing Users functionality


snyde664:
### Why is this issue important?
This is a part of the [User Management User Story](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=50424619).

A task within this story is Adding Users. Building out the adding users functionality is a necessary element of the CPF reporter. Dependent on [this ticket](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=44762614) and [this ticket](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=44763683).

### Current State

None

### Expected State

This functionality should only be accessible to admin users. Clicking on the “Edit” button in a row of the User table leads to the selected User’s page. This is the same page as the Create New User page, but all of the User’s existing information will be displayed:

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/44246589/ecfe0961-8283-4ff0-abd9-d86e01c5e598)

Here, the user should be able to edit all the information beside the ID number, and click the Save button to save their changes. These changes should be reflected in the table of users.

### Implementation Plan

Same as #15 

Data-model:
https://lucid.app/lucidchart/93e8fea1-fd5b-4722-901b-cfe6703f3798/edit?invitationId=inv_ca230f47-f1bf-4594-a959-696293102dbf&page=0_0#

CRUD page creation:
Cells to view data: https://redwoodjs.com/docs/tutorial/chapter2/cells
Forms to accept data: https://redwoodjs.com/docs/tutorial/chapter3/forms

as1729:
Recommend to work on #15  together with this ticket.

ClaireValdivia:
This is working great in staging

Issue #15, opened 2023-11-15
User Management: Adding Users functionality


snyde664:
### Why is this issue important?
This is a part of the [User Management User Story](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=50424619).
One of the tasks within this user story is Adding Users. Building out the adding users functionality is a necessary element of the CPF reporter. Dependent on [this ticket](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=44762614).

### Current State

None

### Expected State

This functionality should only be accessible to admin users. Clicking on the “Create New User” button should lead to the Create New User page:

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/44246589/55fecdb2-4485-4cd5-94c2-81e1130548de)

This page includes the following:

- Page title of “User”
- Frozen ID textbox
- Email textbox
- Name textbox
- Role dropdown with 2 options: staff or admin
- Agency dropdown with existing agencies in the tenant as options
- Reset button: clears all entered information
- Save button: adds a user based on entered information
  - Email, name, role, and agency are required fields. If any of these is missing, an error message will appear under the missing field textboxes and the user will not be saved.

Saving a user adds them to the database, and therefore the table, of users and creates a user account for them. This allows the user to login with their associated email.

### Implementation Plan

Data-model: 
https://lucid.app/lucidchart/93e8fea1-fd5b-4722-901b-cfe6703f3798/edit?invitationId=inv_ca230f47-f1bf-4594-a959-696293102dbf&page=0_0#

CRUD page creation:
Cells to view data: https://redwoodjs.com/docs/tutorial/chapter2/cells
Forms to accept data: https://redwoodjs.com/docs/tutorial/chapter3/forms

as1729:
Recommend to work on #16 together with this ticket.

ClaireValdivia:
This is working great in staging. Closing ticket!

Issue #14, opened 2023-11-15
User Management: Users page / Viewing Users functionality


snyde664:
### Why is this issue important?
This is a part of the [User Management User Story](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=50424619).
This work takes place primarily under the Users tab of the webapp. The first step to building this functionality is building out the users page and table, which allows for the Viewing Users functionality.

### Current State

None

### Expected State

The Users tab and associated functionality should only be accessible to admin users. Clicking on the Users tab should bring the user to the Users page. In the Users page, they should see a table of all current users in their tenant’s organization: 

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/44246589/6d93129a-059f-4d13-93bb-ebc3eb318419)

This page should include:

- page title of “Users”
- “Create New User” Button [this ticket does not include New User implementation, the button just needs to be made. New User Implementation is part of [this ticket](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=44763683)]
- Table of existing users
  - Search capability
  - Columns:
  - Email
  - Name
  - Role: Staff or Admin
  - Agency
  - Created: Creation time / date
  - Edit: Edit button for individual user [this ticket does not include edit implementation, the button just needs to be made. edit button implementation is on [this ticket](https://github.com/orgs/usdigitalresponse/projects/5/views/34?pane=issue&itemId=44764610)]

The user should be able to sort the table by each column. Users should be able to use the search capability or sort by each column to find users.

### Implementation Plan

Data-model: 
https://lucid.app/lucidchart/93e8fea1-fd5b-4722-901b-cfe6703f3798/edit?invitationId=inv_ca230f47-f1bf-4594-a959-696293102dbf&page=0_0#

CRUD page creation:
Cells to view data: https://redwoodjs.com/docs/tutorial/chapter2/cells
Forms to accept data: https://redwoodjs.com/docs/tutorial/chapter3/forms

as1729:
Related to #15 and #16 please work on the three together.

ClaireValdivia:
Tested in staging and this is looking great on the usdr admin role - will create an issue for making the User tab visible to Org Admins

PR #48, opened 2023-12-12
14/View users page, add user page, edit user page

ifranch:
This PR addresses tickets #14 for creating a Users page to view users, #15 to add new users, and #16 to edit existing users.

I followed a similar model to https://github.com/usdigitalresponse/cpf-reporter/pull/27 and followed the decisions made there.

### View users

This shows existing users. For now, it only shows users with hardcoded organization id 1. Additionally, the agency and role are shown as ids and not by name. I'm open to feedback on how to change this for this PR or in the future

Screenshot 2023-12-12 at 3 42 36 PM

### Add new user

As indicated in the ticket, this form does not have the organization id field. For now, users are created without an organization id. The agency and role have to be entered as ids for now since that is the corresponding field in the User object.

Screenshot 2023-12-12 at 3 45 12 PM

### Edit user

This is a similar form to the one to add new users but it also displays the created time and user id, which are both disabled and not editable.

Screenshot 2023-12-12 at 3 45 59 PM

### Testing

Given some of the limitations described above, you can use Prisma Studio to create some of the objects such as Role or to add additional fields that are not yet supported like organization id. You can run the following command to access the interface

`yarn redwood prisma studio`

To test the new features:

`yarn redwood dev`

and go to the users page, the new user page and the edit user page.


dysmento:
I get an error when creating a new user:

> Argument `updatedAt` is missing

I think we need one of these options:

1. make updatedAt optional
2. give updatedAt a default value of now()
3. add updatedAt to the form so a value can be set there

Of these, 2 is my favorite.

ifranch:
> I get an error when creating a new user:
> 
> > Argument `updatedAt` is missing
> 
> I think we need one of these options:
> 
> 1. make updatedAt optional
> 2. give updatedAt a default value of now()
> 3. add updatedAt to the form so a value can be set there
> 
> Of these, 2 is my favorite.

This is already fixed in this PR. I was getting that error too initially and then I changed the user schema to add the @updatedAt @db.Timestamptz(6) notation to the updatedAt field, which is similar to the option 2 you suggested. After making the change, I was able to add new users without issues and that field was auto-populated. Can you check that the schema.prisma file is properly updated in the branch you are testing?

as1729:
@ifranch thanks for submitting the PR! are you able to resolve the conflicts with the `main` branch here and merge? I don't have any blockers on my end here to merge this code.

as1729 (approved):

as1729 (approved):

PR #49, opened 2023-12-13
23 adding localstack config, docker-compose.yml and the aws library w…

ed-snodgrass:
I created aws.ts based off of the gost-aws.js from the other project. I moved some of the details for things like writing to s3 and getting signed urls so that all that needs to happen is a bucket/key needs to be provided (and an object when storing). I left out the SES stuff because it didn't seem to be used in the other project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,5 @@ dist
 # Terraform cache
 terraform/.terraform
 .tool-versions
+
+localstack/volume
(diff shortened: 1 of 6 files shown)


dysmento (approved):
I checked out this branch locally, ran docker compose up on the localstack directory, and initialized successfully with `localhost/entrypoint/init-aws.sh`

c.f. I installed awslocal like this:

`pip3 install awscli-local[ver2]`

as1729 (approved):

PR #51, opened 2023-12-14
Add Dashboard page (Uploads table), create a reusable TableBuilder

Vikariusu:
# Ticket #3 

This PR creates a dashboard page that renders Uploads table. I also took a stab at adding filters and sorting functionality, resulting in a reusable `<TableBuilder />` component. This uses the most popular table library (tanstack table), which may prove beneficial when adding more functionality down the line.

### Page
- [x] Renders on "/" route
- [x] Contains Uploads heading
- [x] 3 buttons below the heading match the wording with the design on Figma

### Table
- [x] Contains columns: ID, Agency, EC Code, Uploaded By, Filename, Validated?
    - doesn't contain column "Notes" as this field is not present in the Upload model
- [x] Columns are sortable
    - the only not-sortable column is the validation column as it's a date field that would require additional logic
 - [x] 4 fields have option to filter (as a textfield) - Agency, EC Code, Uploaded By, Filename
     - ID doesn't have a filter option as per Figma design
     - Validated? doesn't have a filter option, as filters are now just textfields. Validated is a dropdown on Figma and requires additional logic
 - [x] Table contains "Reset filters" button

## TableBuilder and TanStack Table
Since we'll probably want to add filters, sorting and possibly pagination in the future, I went ahead and implemented a reusable `<TableBuilder />` component made using TanStack Table (~1 mln downloads weekly).

`<TableBuilder />` includes sorting and filtering (optional). 

## Styling
The table uses Bootstrap, the design in Figma recreates an existing ARPA table that uses Vue Good Table (obviously not an option for this project).

https://github.com/usdigitalresponse/cpf-reporter/assets/20152824/5995d057-e378-4053-b7e0-a89c6630c86d


web/src/components/Upload/Uploads/Uploads.tsx
```diff
@@ -0,0 +1,65 @@
...
+import type { FindUploads } from 'types/graphql'
+
+import TableBuilder from 'src/components/TableBuilder/TableBuilder'
+
+import { columnDefs } from './columns'
+
+const UploadsList = ({ uploads }: FindUploads) => {
+  // Uploads.validations is an array that can contain both valid and invalid validations,
+  // We store the latest valid and invalid entry
+  function findLatestValidationAndInvalidationDatesFromUpload(upload) {
```
Vikariusu, line 12: This code could be a lot cleaner, ideally we'd want to pass `uploads` straight into `<TableBuilder />` (or maybe we could process `validations` on the backend?).

In ARPA, uploads have 2 date strings for validated and invalidated dates, I tried to replicate that here from the `validations` array by getting the latest possible date value (one for the invalidation date, one for the validation one), does this look right? @as1729

web/src/components/Upload/Uploads/Uploads.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (!date1) return date2
+    if (!date2) return date1
+    return date1 > date2 ? date1 : date2
+  }
+
+  /*
+    Returns an object that either contains the latest validation date or the latest invalidation date
+    (depending which validation date is later) or an empty object if no dates are found
+  */
+  function findLatestValidationEntry(upload) {
```
Vikariusu, line 24: Updated the logic around upload validations. This function checks both for `invalidatedAt` and `validatedAt` dates and determines which one was the latest (or returns an empty object if no dates are set).

It might be easier to condense this data and replace `invalidatedAt` and `validatedAt` (as well as `invalidatedBy` and `validatedBy`) with something like `reviewedAt` (and `reviewedBy`) and then have a boolean `isValid`. This would simplify the frontend logic considerably and reduce the complexity in the `UploadValidation` model.

web/src/components/Upload/Uploads/Uploads.tsx
as1729: Confirmed via Slack that we will use the latest possible date value.

web/src/components/Upload/Uploads/Uploads.tsx
as1729: 100% great idea. Lmk if you would prefer to merge the PR as-is and tackle this model change separately.

as1729 (approved):

PR #53, opened 2023-12-17
Feature/24 submit workbook page

dysmento:
This is work in progress. I wanted to see how straightforward it would be to copy `validate-upload.js` over from ARPA reporter. The answer is not very, because we also need `validation-rules.js` `ec-codes.js` `records.js` `validation-error.js` `persist-upload.js` `templateRules.json` and maybe a couple I'm not remembering.

As much as I could, I brought these files over with a `.ts` extension, but in some cases I got a lot of type errors I didn't understand, so I fell back to `.js` in those cases.

Questions that came up while I worked on this:
1. does this approach even make sense? I don't know if copying a bunch of code like this can be made to work in a different language on a different framework. The alternative would be to make small changes bit by bit, like a single `validateUpload()` function with a single rule like maybe `validateVersion` and move on from there.
2. I somehow got the impression we were storing uploads in S3, but now I'm not sure, because ARPA reporter uses a filesystem. Where are we storing uploads?
3. I saw a note in the lucid ERD diagram: "Is it worth having a separate validations table?" And now I'm wondering that too. It makes sense if you want to keep a history of every validation (and invalidation). Is that what we're going for?
4. I notice the schema has no Notes field for an upload. Did we jettison the notes field?
5. ARPA code has a way to look up a recipient by tax ID number, but CPF schema doesn't have one. Do we not need that?
6. In much of the ARPA code, we're passing around a knex transaction object. I don't think prisma lets you access the transaction in the same way. What's the right approach to take for functions that need to participate in a prisma transaction?

I think someone should take this over from me, because my availability is not great until Friday. TODO:
1. I added the form fields we'll need for upload workbook, but we still need to read from the database for the organization's list of reporting periods and agency codes
2. I got an error trying to require `api/src/lib/templateRules.json` and I tried a couple of different ways but I couldn't figure it out. It's needed by `validation-rules.js`
3. the file data needs to be persisted
4. I added a call to `validateUpload` to `api/src/services/uploads/uploads.ts` but there's no way it works yet, and it needs to be made to work.
5. I didn't have a chance to start creating the view that shows success or failure of the validation.

I almost forgot to mention: none of the tests have been brought over yet, which is probably not a small job, since we'd be moving them from mocha to jest.


web/src/components/Upload/UploadForm/UploadForm.tsx
```diff
@@ -21,135 +25,73 @@ interface UploadFormProps {
...
+        name="reportingPeriodId"
+        className="rw-label"
+        errorClassName="rw-label rw-label-error"
+      >
+        Reporting Period
+      </Label>
+      <SelectField name="reportingPeriodId">
+        <option value={1}>23Q3</option>
+        <option value={2}>23Q4</option>
+        <option value={3}>24Q1</option>
```
dysmento, line 62: We're going to need a variant of an Organization Cell that takes an organizationId and returns the agency and reporting period options to use here.

api/src/lib/records.js
```diff
@@ -0,0 +1,298 @@
...
+//       [
+//         ...Object.values(EC_SHEET_TYPES),
+//         'awards50k',
+//         'expenditures50k',
+//       ].includes(record.type)
+//     )
+
+//   return Object.values(projectRecords)
+// }
+
```
dysmento, line 291: I commented out a fair amount of code because it isn't needed immediately for validating uploads.

api/src/lib/validate-upload.js
```diff
@@ -0,0 +1,793 @@
...
+import {
+  subrecipient,
+  createSubrecipient,
+  updateSubrecipient,
+} from 'src/services/subrecipients'
+import {
+  updateUpload,
+  markValidated,
+  markInvalidated,
+} from 'src/services/uploads'
```
ed-snodgrass, line 14: there isn't a file at src/services/uploads. when I updated this import to point to `src/services/uploads/uploads` it was missing the markValidated and markInvalidated functions. Also, that uploads file is importing validate-upload, so I expect that circular dependency is causing an issue.

ed-snodgrass:
@dysmento , regarding your question about S3, the intent here is to use S3 for all files, while, I believe, ARPA does upload some files to a filesystem and some to S3.

as1729:
Discrete next tasks

- Upload to S3 and kick-off validation
- Validation processing (Riskiest piece)
  - Sample json needed - of both the templateRules.json as well as the upload validation
- Results page

ed-snodgrass (approved):
Looks good

Vikariusu (approved):

PR #54, opened 2023-12-18
CPF-34 Add Navigation layout, use private routes

Vikariusu:
# Ticket #34 

## Navigation
- Created a navigation using Bootstrap tabs
- "Subrecipients" and "Users" tabs are disabled as they require the pages to be complete. Otherwise links that use Router would break the application.
- Added an "Organizations" tab at the end as consulted with Aditya (later to be conditional depending on a specific user role)

## Authenticated routes
All pages were wrapped in the `<PrivateSet />`. I'm not currently able to access `currentUser` coming from the `useAuth()` hook, so I commented out a conditional logic that will use `currentUser` when the authentication is implemented.

Screenshot 2023-12-17 at 10 09 50 PM


as1729:
@Vikariusu I'm good for this to merge, however, I noticed that the `localstack/` folder got committed. Are you able to update the `.gitignore` [file](https://github.com/usdigitalresponse/cpf-reporter/blob/main/.gitignore#L148) to remove all files in the `localstack/` directory and not just `localstack/volume` moving forward.

Vikariusu:
> @Vikariusu I'm good for this to merge, however, I noticed that the `localstack/` folder got committed. Are you able to update the `.gitignore` [file](https://github.com/usdigitalresponse/cpf-reporter/blob/main/.gitignore#L148) to remove all files in the `localstack/` directory and not just `localstack/volume` moving forward.

@as1729 Good catch, done

localstack/docker-compose.yml
as1729, line 1: @Vikariusu looks like we will need `localstack/docker-compose.yml` and `localstack/entrypoint/init-aws.sh`

Maybe we can try changing `localstack/` to `localstack/volume/`? I wonder if the ending `/` was the reason the previous volume files got committed.

Vikariusu: Not sure, in my IDE the volume folder is grayed out with or without the `/` at the end. But I can try making that change

as1729 (approved):

PR #63, opened 2023-12-25
61 workbookValidation of "version"

ed-snodgrass:


api/src/lib/validate-upload.ts
ed-snodgrass, line 1: @as1729 this is the file that has the markValidated and markInvalidated functions (that it cannot find).

Issue #56, opened 2023-12-21
Organization Setup


snyde664:
### Why is this issue important?

We need rto be able to create organizations to onboard our partners. 

### Current State

none

### Expected State

Need:

- concept of a USDR super-admin in code

- Update the “create organization” action to be scoped for the user-super-admin user
    - creating an organization should tie in with creating an agency and associated admin user

Vikariusu:
## PR info
PR link: https://github.com/usdigitalresponse/cpf-reporter/pull/64
The PR for this ticket has been merged and is currently in staging.

## Functionality Overview

### Organizations tab
1. Organizations tab renders only for users with `USDR_ADMIN` role.
2. For users without `USDR_ADMIN` role, navigating to `/organizations` will show a newly added Forbidden page. Forbidden page renders when the current user doesn't have permissions to view that page but is logged in.

### New Organization form
1. Creates a new organization, a new agency under that organization, and a new user with `ORGANIZATION_ADMIN` role assigned to that organization and agency.
2. All fields are required. Trying to submit a form with empty fields will render "This field is required" under the missing fields.
3. Replaced instances of the term 'Team' with 'Organization' in labels for consistency.

## Dev Changes

### Changing `Role` model to an enum
Made a few changes after talking to Aditya

1. Removed the `Role` model and replaced it with an `enum`. Enums are commonly written in uppercase letters (e.g. `USDR_ADMIN`), which is the convention I've followed.
2.  Kept the hardcoded `roles` array in `getCurrentUser()` but changed the string that we're passing to match the enum formatting.
3. Updated the user adding/editing forms to use a select input (dropdown) for the `Role` field, replacing the previous text field for `Role ID`.
4. Used an existing `formatEnum()` function to change how we display `user.role` on the frontend (eg. `ORGANIZATION_ADMIN` to `Organization admin` )

### New Organization form as a separate component
New Organization form has its own component, separate from Edit Organization form. By default, Redwood uses the same form for adding and editing but since those 2 forms are quite different, separating the two makes the code more clear.

PR #64, opened 2023-12-29
Feature/CPF-56 Organization Creation by USDR admin role

Vikariusu:
# Ticket #56 

This is a draft PR for creating organization form for USDR admin users. There are couple open questions I have to make this feature complete.

Currently we have the Role and User models in `schema.prisma` below:
```prisma
model Role {
  id        Int      @id @default(autoincrement())
  name      String
  createdAt DateTime @default(now())
  updatedAt DateTime
  users     User[]
}

model User {
  id             Int                 @id @default(autoincrement())
  email          String
  name           String?
  agencyId       Int?
  organizationId Int?
  roleId         Int?
  createdAt      DateTime            @default(now()) @db.Timestamptz(6)
  updatedAt      DateTime            @default(now()) @db.Timestamptz(6)
  agency         Agency?             @relation(fields: [agencyId], references: [id])
  organization   Organization?       @relation(fields: [organizationId], references: [id])
  role           Role?               @relation(fields: [roleId], references: [id])
  certified      ReportingPeriod[]
  uploaded       Upload[]
  validated      UploadValidation[] @relation("ValidatedUploads")
  invalidated    UploadValidation[] @relation("InvalidatedUploads")
  subrecipients  Subrecipient[]
}
```

which means that user can only have 1 role. However, the `getCurrentUser()` function that provides (currently hardcoded) user data, uses `roles` array. 

Questions:
1. Can user have more than one role? Which format will be used?
2. If user can only have one role (like the database is currently set up), is there a need for `Role` to be a separate model and not an `Enum`? 
3. The Lucid chart currently mentions 3 different user roles, written in snake case (like `usdr_admin`). Is that the convention we'll be using going forward? Currently I'm using `usdr-admin`.
4. When creating a new admin user for the organization, what is the type of the admin that the user should have? Lucid chart lists `organization_admin` role for example, is this the role that should be assigned?

### Organizations tab
1. Organizations tab renders only for users with `usdr-admin` role.
2. For users without `usdr-admin` role, navigating to `/organizations` will show a newly added Forbidden page. Forbidden page renders when the current user doesn't have permissions to view that page but is logged in.

### New Organization form
1. Creates a new organization, assigns a new agency to that organization, and assigns a new admin user to both the organization and the agency.
2. All fields are required. Trying to submit a form with empty fields will render "This field is required" under the missing fields.
3. The backend logic assumes that the admin role exists and does not create a new role in the database.

## New Organization form as a separate component
New Organization form has its own component, separate from Edit Organization form. By default, Redwood uses the same form for adding and editing but since those 2 forms are quite different, separating the two makes the code more clear.


as1729:
> Can user have more than one role? Which format will be used?

User can only have one role.

> If user can only have one role (like the database is currently set up), is there a need for Role to be a separate model and not an Enum?

Good point. If the implementation is easier, then lets opt to keeping this as an enum on the users table.

> The Lucid chart currently mentions 3 different user roles, written in snake case (like usdr_admin). Is that the convention we'll be using going forward? Currently I'm using usdr-admin.

No preference here. `usdr-admin` or `usdr_admin` are equally acceptable.

> When creating a new admin user for the organization, what is the type of the admin that the user should have? Lucid chart lists organization_admin role for example, is this the role that should be assigned?

Yes, for all users the role should `organization_admin` however, for USDR users, we will manually write a script that creates them.

Vikariusu:
Made a few changes after talking to @as1729 

1. Removed the `Role` model and replaced it with an `enum`. Enums are commonly written in uppercase letters (e.g. `USDR_ADMIN`), which is the convention I've followed.
2.  Kept the hardcoded `roles` array in `getCurrentUser()` but changed the string that we're passing to match the enum formatting.
3. Updated the user adding/editing forms to use a select input for the `Role` field, replacing the previous text field for `Role ID`.
4. Used an existing `formatEnum()` function to change how we display `user.role` on the frontend (eg. `ORGANIZATION_ADMIN` to `Organization admin` )
5. Replaced instances of the term 'Team' with 'Organization' in labels for consistency.

dysmento (approved):
Thanks!

PR #66, opened 2024-01-05
add Lambda function to respond to S3 put events and add json files

dysmento:
Since the backend is not responsible for putting uploads into S3 (that is, it supplies a signed URL, and the frontend is what does the actual S3 Put), I think it makes sense to generate the JSON files using a Lambda function that responds to the S3 Put event.

I have added a function to do that. A couple of things are missing:

1.  we need a bit of terraform to configure the bucket to emit these events and set up the Lambda
2. we still need something to initiate a validation. Should that be in this function, or should it be separate?
3. can the UI have some kind of indicator that validation is still in progress? How do we notify when we're done?


as1729 (approved):
Verified terraform plan is succeeding. Going to test this in staging.

as1729 (approved):

PR #68, opened 2024-01-09
Fix/66 new lambdas iac

TylerHendrickson:
This PR provides fixes and related changes that address issues preventing successful deployment following #66.

Summary of changes:
- Consolidate multiple `aws_s3_bucket_notification` resources for the same bucket in Terraform, per [this note from the Terraform documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_notification)
  > S3 Buckets only support a single notification configuration. Declaring multiple aws_s3_bucket_notification resources to the same S3 Bucket will cause a perpetual difference in configuration. See the example "Trigger multiple Lambda functions" for an option.

  To be clear "perpetual difference" means that the resources will be in conflict, and will create a race condition during `terraform apply` wherein one of the resources will always overwrite the other(s).
- Update the naming convention used to refer to the bucket where CPF Reporter data (uploads and their derived JSON files, as well as reporting templates, and other S3 data managed at runtime) is stored. Since this bucket is used to store more than just uploads, I updated names like `cpf_uploads_bucket` to `reporting_data_bucket`. 
- Reorder and add `// Label` comments to the various inputs on the `lambda_function-graphql` Terraform module, and update the other `lambda_function-` modules (added in #66) to match.
  - Hopefully, this change will contribute to more of a "living example" out of these implementations and make it easier to add new Lambda functions in the future.
  - As discussed with @as1729, more formal documentation around how to add a Lambda function may ultimately be warranted, but this seems like a good starting point. Curious to get @dysmento's thoughts on whether this makes our Lambda conventions more obvious.
  - Would be good to discuss if anything still seems unclear / if there are questions about why a convention exists.
  - NB: In addition to more documentation, I'm also interested in leveraging something like TFLint or another static analysis tool to make it easier to validate conventions in CI workflows at PR time, which would help avoid post-merge failures when deploying to Staging.
- Expose the name of the reporting data bucket as a `$REPORTING_DATA_BUCKET_NAME` environment variable to all Lambda functions.

### DIFF
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -15,7 +15,7 @@ import { getSignedUrl as awsGetSignedUrl } from '@aws-sdk/s3-request-presigner'
 import { StreamingBlobPayloadInputTypes } from '@smithy/types'
 import { QueryResolvers, CreateUploadInput } from 'types/graphql'
 
-const CPF_REPORTER_BUCKET_NAME = `cpf-reporter-${process.env.environment}`
+const REPORTING_DATA_BUCKET_NAME = `${process.env.REPORTING_DATA_BUCKET_NAME}`
 
 function getS3Client() {
   let s3: S3Client
@@ -64,8 +64,8 @@ export function uploadWorkbook(
   uploadId: number,
   body: StreamingBlobPayloadInputTypes
 ) {
-  const folderName = `${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/uploads/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
-  return sendPutObjectToS3Bucket(CPF_REPORTER_BUCKET_NAME, folderName, body)
+  const folderName = `uploads/${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
+  return sendPutObjectToS3Bucket(REPORTING_DATA_BUCKET_NAME, folderName, body)
 }
 
 async function sendPutObjectToS3Bucket(
@@ -85,7 +85,7 @@ async function sendPutObjectToS3Bucket(
 
 export function getTemplateRules(inputTemplateId: number) {
   return sendHeadObjectToS3Bucket(
-    CPF_REPORTER_BUCKET_NAME,
+    REPORTING_DATA_BUCKET_NAME,
     `templates/input_templates/${inputTemplateId}/rules/`
   )
 }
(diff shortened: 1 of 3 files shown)


api/src/lib/aws.ts
```diff
@@ -64,8 +64,8 @@ export function uploadWorkbook(
   uploadId: number,
   body: StreamingBlobPayloadInputTypes
 ) {
-  const folderName = `${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/uploads/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
-  return sendPutObjectToS3Bucket(CPF_REPORTER_BUCKET_NAME, folderName, body)
+  const folderName = `uploads/${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
```
TylerHendrickson, line 67: Given that this path structure is duplicated below, it might be useful to add some kind of helper method/function that returns a string S3 key for a given Upload object, and perhaps another one that returns the S3 for the JSON file resulting from the XLSM-to-JSON conversion.

api/src/lib/aws.ts
```diff
@@ -15,7 +15,7 @@ import { getSignedUrl as awsGetSignedUrl } from '@aws-sdk/s3-request-presigner'
 import { StreamingBlobPayloadInputTypes } from '@smithy/types'
 import { QueryResolvers, CreateUploadInput } from 'types/graphql'
 
-const CPF_REPORTER_BUCKET_NAME = `cpf-reporter-${process.env.environment}`
+const REPORTING_DATA_BUCKET_NAME = `${process.env.REPORTING_DATA_BUCKET_NAME}`
```
TylerHendrickson, line 18: I'm wondering if this actually needs to be a global constant – the only time I can think of where it would actually be necessary to reference a bucket name like this would be at the point where a pre-signed URL was being generated. For any Lambda function that's invoked via a bucket notification, it seems like it would be better to just refer to the invocation event payload.

terraform/functions.tf
```diff
@@ -286,54 +291,130 @@ module "lambda_function-excelToJson" {
...
-  function_name = "excel-to-json"
+  // Metadata
+  function_name = "${var.namespace}-excelToJson"
   description   = "Reacts to S3 events and converts Excel files to JSON."
 
-  vpc_subnet_ids = local.private_subnet_ids
-  vpc_security_group_ids = [
-    module.lambda_security_group.id,
-    module.postgres.security_group_id,
-  ]
```
TylerHendrickson, line 296: Since the `excelToJson` handler doesn't seem to require DB access at the moment, we can avoid adding the Lambda function to the VPC.

terraform/functions.tf
```diff
@@ -286,54 +291,130 @@ module "lambda_function-excelToJson" {
...
+
+  // Metadata
+  function_name = "${var.namespace}-cpfValidation"
   description   = "Reacts to S3 events and validates CPF JSON files."
 
-  vpc_subnet_ids = local.private_subnet_ids
-  vpc_security_group_ids = [
-    module.lambda_security_group.id,
-    module.postgres.security_group_id,
-  ]
```
TylerHendrickson, line 323: Since the `cpfValidation ` handler doesn't seem to require DB access at the moment, we can avoid adding the Lambda function to the VPC.

TylerHendrickson:
@dysmento A few questions related to this set of work:
1. There's a [comment here](https://github.com/usdigitalresponse/cpf-reporter/blob/373c87394217fb9808654821f5d7701513a00cfb/api/src/functions/cpfValidation/cpfValidation.ts#L21-L22) about calling a separate API endpoint for validation. Any reason this would be preferable to just executing the validation logic directly within that Lambda function?
2. If there is a need to do this via an API call, does that endpoint exist yet?
3. I'm seeing some code in `api/src/lib/persist-upload.js` that seems geared to persist files to disk. I'm assuming that's placeholder code for the moment, but can you confirm whether that's the case?

dysmento (approved):
Thanks, Ty. I don't think I could have arrived at the networking and permissions changes on my own.

PR #70, opened 2024-01-11
fix: add auth-provider to allow header

as1729:
Based on the discussion in Slack it appears that the issues we're seeing in staging is due to the lack of `auth-provider` being marked as an allowed header.

image


dysmento (approved):
Seems straightforward. Thanks!

PR #71, opened 2024-01-11
fix: remove duplicate routes

as1729:
Previously we copied a few routes from private-set into unauthenticated. However, this does not provide the appropriate data needed to support the upload page load. Hence this PR does:
1. Remove the duplicate routes
2. Ensure all routes (except login and not found) are within the private-set


dysmento (approved):
ok.

Issue #92, opened 2024-01-27
Upload flow - ensure signedUrl is passed back via createUpload graphql request


as1729:

Vikariusu:
The PR for this ticket was merged on 02/06.

Copying the PR's description:
## Description
We're now passing back `signedUrl` along with the newly created upload in the `createUpload` resolver.

## Testing
I wasn't able to run localstack, so part of the signedUrl contains `undefined`. Before reviewing this PR, make sure you have localstack running and are able to test this feature.

![signedurl](https://github.com/usdigitalresponse/cpf-reporter/assets/20152824/b494b6c7-6f63-4b96-82cc-5a083783f003)

PR #96, opened 2024-02-05
CPF-92 Pass back signedUrl when creating a new upload

Vikariusu:
# Ticket #92 

## Description
We're now passing back `signedUrl` along with the newly created upload in the `createUpload` resolver.

## Testing
I wasn't able to run localstack, so part of the signedUrl contains `undefined`. Before reviewing this PR, make sure you have localstack running and are able to test this feature.

signedurl


as1729 (approved):
Even though localstack isn't working correctly the fact that the URL is being passed back is sufficient validation that this particular issue is resolved. Will test in staging to see if there are any issues.

PR #100, opened 2024-02-07
Fix: S3 pre-signed uploads (PR #98 follow up)

TylerHendrickson:
This PR fixes two issues with pre-signed S3 URLs:
1. A CORS misconfiguration, which currently omits `https://` from the allowed origin (website) domain name. This should have been included in #98.
2. An issue with the `PutObject` command used to configure pre-signed URLs for web-side upload. Currently, the command is configured without AES256 server-side encryption, which is a [requirement](https://github.com/usdigitalresponse/cpf-reporter/blob/4cf3d15cbdd6422d676ebbaac1b10fe49daf4803/terraform/functions.tf#L129-L131) of this S3 bucket. Without this fix, object uploads to S3 (regardless of CORS configuration) will continue to fail.

### DIFF
--- a/terraform/functions.tf
+++ b/terraform/functions.tf
@@ -134,7 +134,7 @@ module "reporting_data_bucket" {
   cors_configuration = [
     {
       allowed_methods = ["GET", "HEAD", "POST", "PUT"]
-      allowed_origins = [var.website_domain_name]
+      allowed_origins = ["https://${var.website_domain_name}"]
       allowed_headers = ["*"]
       expose_headers = [
         "Content-Disposition",
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -110,6 +110,7 @@ export async function s3PutSignedUrl(
     Key: key,
     ContentType:
       'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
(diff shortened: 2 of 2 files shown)


terraform/functions.tf
```diff
@@ -134,7 +134,7 @@ module "reporting_data_bucket" {
   cors_configuration = [
     {
       allowed_methods = ["GET", "HEAD", "POST", "PUT"]
```
as1729, line 136: should we add `OPTIONS` as well @TylerHendrickson ?

TylerHendrickson: Generally no, `OPTIONS` shouldn't be necessary (well-noted that our API Gateway configuration does specify `OPTIONS`) for CORS configuration.

[This StackOverflow answer](https://stackoverflow.com/a/68529748) goes a bit deeper into the reason why, but I think the short version is just that a server doesn't need to explicitly allow `OPTIONS` requests because that's the mechanism by which allowed methods are determined in the first place. Including `OPTIONS` assumes that a browser would make a preflight (`OPTIONS` method) request with an `Access-Control-Request-Method: OPTIONS` header, which is a bit like calling someone on the phone to ask if you are allowed to call them on the phone.

as1729 (approved):
🚀

PR #108, opened 2024-02-14
Adds ability for developers to authenticate users in the local environment

as1729:
## Context
With the implementation of [Passage](docs.passage.id) we are introducing a dependence on a third-party service to manage user-authentication. This can pose a challenge for local development if every developer needs to sign-up to get API-keys etc.
This PR ensures that in the local environment developers have the ability to authenticate a user merely by providing an email address of a valid user.

## How does the flow of data work?
1. `LoginPage` front-end component renders the appropriate page depending on whether the environment's auth-provider is `local | custom-auth | passage`. Note: `custom-auth` will be deprecated once `passage` is implemented.
image
image

2. `web/src/auth.ts` instantiates the correct auth implementation client based on the same `AUTH_PROVIDER` environment variable.

3. The Auth client makes requests to `localhost:8911/localAuth` which is a new endpoint served by `api/src/functions/localAuth/localAuth.ts` function.

4. The `localAuth.ts` handler verifies the request and looks for the user-email submitted and checks whether a user with that email exists in the database. If the user is not found it sends a 4xx code or sends the user-object back in the 200 response.

5. The front-end stores the email the `local_auth_token` key of `localStorage`. This will then be passed in as part of the header in subsequent `localhost:8911/graphql` requests as `Authorization: Bearer ${localStorage.local_auth_token}`

6. The `createGraphQLHandler` then passes this information to the `getCurrentUser` function in  `api/src/lib/auth` which does a similar check as Step 4 and passes back the user that is tied to the token passed in the `Authorization` header.

## Testing
1. Ensure the database is seeded with the latest users added in this PR.
2. Ensure that the `AUTH_PROVIDER` environment variable is set to `local` and the app has restarted since changing this variable.
3. Navigate to `localhost:8910/login`
4. Verify that `foo@example.com` redirects back to the login page.
5. Verify that `grants-admin@usdigitalresponse.org` successfully redirects to `localhost:8910/`
6. Verify that hitting `Logout` on the top right redirects back to `/login`
7. Verify that going to root `/` redirects back to `/login` since no user is logged in


web/src/auth/localAuth.ts
```diff
@@ -0,0 +1,118 @@
...
+  },
+  getUserMetadata: async () => {
+    if (!localStorage.getItem('local_auth_token')) {
+      return null
+    }
+
+    console.log('local: getUserMetadata')
+    let user: User | null = null
+    // Explicitly setting the url to ensure these requests are not made outside of the local environment
+    await fetch(`http://localhost:8911/localAuth`, {
```
TylerHendrickson, line 52: See the related comment in the API-side `auth.ts` – I think the env var should be good enough to dictate whether local auth is enabled; there are plenty of other things that could go awry if we aren't careful about env vars in hosted environments.

web/src/auth/localAuth.ts
```diff
@@ -0,0 +1,118 @@
...
+        user = jsonData
+      })
+      .catch((error) => {
+        console.log(error.status)
+        throw new Error('Error getting user metadata')
+      })
+    return user
+  },
+  login: async (event) => {
+    await fetch(`http://localhost:8911/localAuth`, {
```
TylerHendrickson, line 78: Same as the other two related comments

web/src/auth/localAuth.ts
```diff
@@ -0,0 +1,118 @@
...
+        return Promise.reject(response)
+      })
+      .catch((error) => {
+        console.log(error.status)
+        throw new Error('Error getting user metadata')
+      })
+
+    localStorage.setItem('local_auth_token', event.email)
+    window.location.href = '/'
+  },
```
TylerHendrickson, line 101: As discussed in Slack, I'm wondering if this even needs to hit the API or if it can simply set the `event.email` as the token in local storage without making any service calls.

TylerHendrickson (changes requested):
A few preliminary suggestions. As discussed in Slack, I think it would be ideal to implement a "local auth" option using the minimal custom auth interface provided by Redwood. In addition to lowering complexity, I think it would more closely following the auth implementation that we'll use for Passage. 

And just to add – if that doesn't seem feasible (since I'm admittedly only relying on "in theory it should work based on the docs" here), it would definitely be useful to better understand the caveats.

scripts/seed.ts
```diff
@@ -8,45 +8,45 @@ export default async () => {
...
+    const users: Prisma.UserCreateArgs['data'][] = [
       {
-        email: 'ORGANIZATIONSTAFF@testemail.test',
-        name: 'Organization Staff',
-        role: 'ORGANIZATION_STAFF',
-        agencyId: agency.id,
+        email: 'usdr-admin@usdr.dev',
+        name: 'USDR Admin',
+        agencyId: mainAgencyRecord.id,
+        organizationId: organizationRecord.id,
```
TylerHendrickson, line 31: `organizationId` was removed in #111, which causes this script to fail
```suggestion
```

scripts/seed.ts
```diff
@@ -8,45 +8,45 @@ export default async () => {
...
+        role: 'USDR_ADMIN',
       },
       {
-        email: 'ORGANIZATIONADMIN@testemail.test',
+        email: 'organization-admin@usdr.dev',
         name: 'Organization Admin',
+        agencyId: mainAgencyRecord.id,
+        organizationId: organizationRecord.id,
```
TylerHendrickson, line 38: `organizationId` was removed in #111, which causes this script to fail
```suggestion
```

scripts/seed.ts
```diff
@@ -8,45 +8,45 @@ export default async () => {
...
       },
       {
-        email: 'USDRADMIN@testemail.test',
-        name: 'USDR Admin',
-        role: 'USDR_ADMIN',
-        agencyId: agency.id,
+        email: 'organization-staff@usdr.dev',
+        name: 'Organization Staff',
+        agencyId: mainAgencyRecord.id,
+        organizationId: organizationRecord.id,
```
TylerHendrickson, line 45: `organizationId` was removed in #111, which causes this script to fail
```suggestion
```

scripts/seed.ts
```diff
@@ -8,45 +8,45 @@
...
+        role: 'ORGANIZATION_STAFF',
       },
     ]
-
-    await db.user.createMany({
-      data: users,
-    })
+    const userRecord = await db.user.create({ data: users[0] })
```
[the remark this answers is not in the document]
TylerHendrickson: @as1729 I was going to note this as well – can this line be replaced by moving the create-user operation here instead?
```suggestion
    // Note: if using PostgreSQL, using `createMany` to insert multiple records is much faster
    // @see: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#createmany
    await Promise.all(
      users.map(async (data: Prisma.UserCreateArgs['data']) => {
        const record = await db.user.create({ data })
        console.log(record)
      })
    )
```

scripts/seed.ts
```diff
@@ -66,6 +66,13 @@ export default async () => {
 
     // Note: if using PostgreSQL, using `createMany` to insert multiple records is much faster
     // @see: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#createmany
+    await Promise.all(
+      users.map(async (data: Prisma.UserCreateArgs['data']) => {
+        const record = await db.user.create({ data })
+        console.log(record)
+      })
+    )
+
```
TylerHendrickson, line 75: Per the previous comment, this should be moved closer to where the users are being defined instead of being separated by the `inputTemplates` and `outputTemplates` definitions.
```suggestion
```

TylerHendrickson (changes requested):
Suggesting a few things to fix up. I'm also noticing a few issues when browsing, although they might be out of scope for this issue:
1. Navigating to the `/agencies` route/tab throws an error because something is still looking for `.organizationId` where it doesn't exist.
2. (When logged in as the `usdr-admin@` user) I don't see the "Users" tab in the header menu. If I navigate to `/users` directly, I get the text "Access to this page is restricted."

TylerHendrickson (approved):

PR #109, opened 2024-02-15
chore: remove implementation of login() and getUserMetadata()

as1729:


web/src/auth/localAuth.ts
```diff
@@ -42,59 +42,64 @@ export const localAuthClient = {
     return localStorage.getItem('local_auth_token')
   },
   getUserMetadata: async () => {
-    if (!localStorage.getItem('local_auth_token')) {
-      return null
+    return {
```
as1729, line 45: This function _must_ return a not-null value in order for this to work. So we can either return some valid hard-coded value as seen in this PR or something that looks like the following:
```
{
    email: localStorage.getItem('local_auth_token'),
}
```

Issue #103, opened 2024-02-09
[Issue] User Management: Update `User` Prisma model


TylerHendrickson:
**Subtask of user story:** User Management https://github.com/usdigitalresponse/cpf-reporter/issues/81

**DOD:** The `User` Prisma model is updated to only implement direct relations and make required foreign keys non-nullable. 

**Blocked by issues:** (None)
**Blocks these issues:**
- https://github.com/usdigitalresponse/cpf-reporter/issues/101
- https://github.com/usdigitalresponse/cpf-reporter/issues/102

**Implementation details:**
- Define a `User` Prisma model that implements the following schema:
  - Columns:
    - `id`: Int, pk
    - `email`: string, valid email address, required, add unique
    - `name`: string, required
    - `agencyId`: int, fk to `Agency.id`, **required**
    - `createdAt`: DateTime with timezone, default `now()`, required
    - `updatedAt`: DateTime with timezone, default `now()`, required
    - `role`: value of `Role` enum, required
    - `isActive`: boolean, default `true`, required
  - Relations: 
    - `agency`: 
      - type `Agency` 
      - via `User.agencyId` referencing `Agency.id`
    - `certified`: 
      - type `[]ReportingPeriod` (to-many) 
      - via `ReportingPeriod.certifiedById` referencing `User.id`
    - `uploaded`: 
      - type `[]Upload` (to-many) 
      - via `Upload.uploadedById` referencing `User.id`
- [Seed development fixtures](https://redwoodjs.com/docs/canary/cli-commands#prisma-db-seed) for each user role (and dependencies thereof, e.g. `Agency` and `Organization`) in `scripts/seed.ts`
  - Fixtures that can successfully seed a development environment will suffice in lieu of unit tests for this issue.

as1729:
@TylerHendrickson for valid email validation, do we want to enforce this in the backend and database layer or is it sufficient to handle on the front-end?

PR #111, opened 2024-02-16
CPF-103 Update User model

Vikariusu:
# Ticket #103 

## Description
Updated the `User` model to have all fields required and removed `validated`, `invalidated` and `subrecipients` relations. Tests have been also updated to include all required fields and match changes in the schema.

## Warning
**Changes to the schema break some existing functionality, mainly fetching and creating/updating users.**

### DIFF
--- a/api/db/schema.prisma
+++ b/api/db/schema.prisma
@@ -35,7 +35,7 @@ model Organization {
 
 model User {
   id             Int                 @id @default(autoincrement())
-  email          String
+  email          String              @unique
   name           String
   agencyId       Int
   createdAt      DateTime            @default(now()) @db.Timestamptz(6)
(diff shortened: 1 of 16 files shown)


api/db/schema.prisma
```diff
@@ -37,19 +36,15 @@ model Organization {
 model User {
   id             Int                 @id @default(autoincrement())
   email          String
```
Vikariusu, line 38: Question: Do we want to implement email validation when calling create/update user or is there a way to enforce it here? Also, do we want to use `unique`?

as1729: yes, we would like to use unique here as well 👍 =1

Vikariusu: done ✅

as1729 (approved):
Looks good!

Issue #101, opened 2024-02-09
[Issue] User Management: Create a user via the API


TylerHendrickson:
**Subtask of user story:** User Management https://github.com/usdigitalresponse/cpf-reporter/issues/81

**DOD:** A `createUser` graphQL mutation resolver exists which allows authorized users to create new users via the API

**Blocked by issues:** 
- https://github.com/usdigitalresponse/cpf-reporter/issues/103

**Blocks these issues:** 
- https://github.com/usdigitalresponse/cpf-reporter/issues/106

**Implementation Details:**
- Define a `CreateUserInput` input type in `users.sdl.ts` that implements the following data model, per the `User` Prisma model:
  - `email`: string, required
  - `name`: string, required
  - `agencyId`: int, required, fk to `Agency.id`
  - `role`: enum `RoleEnum`, one of `USDR_ADMIN`, `ORGANIZATION_ADMIN`, `ORGANIZATION_STAFF`
  - `isActive`: boolean, required
-  Ensure that the gql schema in `users.sdl.ts` implements a mutation for `createUser` that:
    - Accepts input conforming to the `CreateUserInput` input schema
    - Uses the `@requireAuth` directive to restrict access to authenticated users with a role of `USDR_ADMIN` or `ORGANIZATION_ADMIN`.
- Define an API-side `users` service with a `createUser` mutation resolver which satisfies the following scenarios:
    - _Note: use [service validations](https://redwoodjs.com/docs/canary/services#service-validations) to validate input and other user creation constraints. Relevant service validations are linked below; in some cases, the gql `CreateUserInput` schema definition will provide validation automatically, in which case service validation may not be necessary_
    - Users may only be created when the following input conditions are satisfied:
      - `email` must be a [valid email address](https://redwoodjs.com/docs/services#validate) ([see also](https://redwoodjs.com/docs/services#email))
      - `email` must be [unique](https://redwoodjs.com/docs/services#validateuniqueness)
      - `name` must [not be null](https://redwoodjs.com/docs/services#presence)
      - `role` must [not be null](https://redwoodjs.com/docs/services#presence) and must be a valid `UserRole` enum
      - `agencyId` must refer to an [existing agency](https://redwoodjs.com/docs/services#validatewith)
    - Requests made by a user with role `USDR_ADMIN` should be able to create a user of any role for any organization
[...]

PR #112, opened 2024-02-19
CPF-101 Update createUser() mutation resolver

Vikariusu:
# Ticket #101

**This PR depends on the schema changes implemented in #111**

## Changes
1. Updated `CreateUserInput` in `users.sdl.ts` to match the requirements (with one exception, see [this](https://github.com/usdigitalresponse/cpf-reporter/pull/112#pullrequestreview-1889237594) comment)
2. Added Redwood service validations in `createUser()` resolver. We already have checks for required fields on the frontend (in the form), and in GraphQL schema.
	1. `Email`: valid email address, unique
	2. `Name`: present, cannot be an empty string (also enforced from gql but the validation here is used so we can display it to the user)
	3. `Agency`: USDR admins can create all users, organization admins can create users within the same organization
3. `createUser()` uses `@requireAuth` directive so only USDR admins and Organization admins can use the mutation (this was implemented in https://github.com/usdigitalresponse/cpf-reporter/pull/95)


api/src/graphql/users.sdl.ts
```diff
@@ -26,9 +26,10 @@ export const schema = gql`
 
   input CreateUserInput {
     email: String!
-    name: String
-    agencyId: Int
-    role: RoleEnum
+    name: String!
+    agencyId: Int!
+    role: RoleEnum!
+    isActive: Boolean
```
Vikariusu, line 32: According to the ticket's requirement `isActive` field on `CreateUserInput` should be required. However, this means that we either need to pass `isActive: true` from the frontend (currently the form for creating users doesn't have this field), or make this field optional on `CreateUserInput`, allowing `schema.prisma` to set it to true by default.
If we would rather set it on the frontend, I can make this change.

as1729 (approved):

PR #128, opened 2024-03-12
Implement access controls for user queries

laurelkorwin:
- Create a common `currentUserIsUSDRAdmin` function to use across queries
- Modify `users`, `user` and `usersByOrganization` queries to use above function and, if user is not USDR admin, limit access only to other users in the querying user's organization
- Unit tests for the above changes

### DIFF
--- a/api/src/services/users/users.test.ts
+++ b/api/src/services/users/users.test.ts
@@ -1,6 +1,14 @@
 import type { User } from '@prisma/client'
 
-import { users, user, createUser, updateUser, deleteUser } from './users'
+import {
+  users,
+  user,
+  createUser,
+  updateUser,
+  deleteUser,
+  currentUserIsUSDRAdmin,
+  usersByOrganization,
+} from './users'
 import type { StandardScenario } from './users.scenarios'
 
 // Generated boilerplate tests do not account for all circumstances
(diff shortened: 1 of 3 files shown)


api/src/services/users/users.test.ts
```diff
@@ -42,6 +112,11 @@ describe('users', () => {
   })
 
   scenario('updates a user', async (scenario: StandardScenario) => {
+    mockCurrentUser({
```
laurelkorwin, line 115: This and `delete` are using `user` under the hood, so they follow the same validation (which I think they should 😉 ). I tested explicitly as admin here to allow the update.

api/src/services/users/users.ts
```diff
@@ -10,13 +11,53 @@ import { AuthenticationError } from '@redwoodjs/graphql-server'
...
+}
+
 export const users: QueryResolvers['users'] = () => {
-  return db.user.findMany()
+  if (currentUserIsUSDRAdmin()) {
+    return db.user.findMany()
+  }
+
+  const userAgency = context.currentUser.agency as Agency
+  validate(userAgency.organizationId, {
```
laurelkorwin, line 24: We can remove all these validations when we make organization non-nullable on agency, see comment here: https://usdigitalresponse.slack.com/archives/C056HNBK12T/p1710271751319379?thread_ts=1710271575.786199&cid=C056HNBK12T

as1729 (approved):

Issue #135, opened 2024-03-22
Agency Report Submission - cpfValidation Lambda function


as1729:
### Why is this issue important?

See User Story here https://github.com/usdigitalresponse/cpf-reporter/issues/73

This is an issue in the sequence of tickets to ensure the CPF reporter can accept a user-submitted upload file and receive feedback on any errors that are present in the submitted file data.

### Blocked By

- [ ] #132 

### Current State

Does not exist

### Expected State

When the [Validation Service](https://gist.github.com/as1729/617a928b18baaa9d003deab3d7e3d07d#file-cpf_validation-py-L396) creates a JSON file containing the results of the validation process there will be an S3 event emitted notifying this service of this object creation.

The goal of this service is to:
1. handle this request and transfer the contents of the JSON into `UploadValidation.results` blob field. The uploadId will be found in the lambda request context.
2. Delete the file from S3 after successfully adding the contents into the database.

### Implementation Plan

1. There is an existing function that already takes in an S3 file creation request. We can re-purpose this function to implement the business logic: https://github.com/usdigitalresponse/cpf-reporter/blob/main/api/src/functions/cpfValidation/cpfValidation.ts
2. Add the following additional code to the handler. Note: this can be separated out into its own function for easier unit testing. This code is added here to explain what logic needs to be implemented.
```ts
import { https } from 'https'

import { S3Event, S3Handler } from 'aws-lambda'

import { logger } from 'src/lib/logger'

import { db } from 'src/lib/db'


/* eslint-disable @typescript-eslint/no-unused-vars */
export const handler: S3Handler = async (event: S3Event): Promise<void> => {
  try {
    const bucket = event.Records[0].s3.bucket.name
    const key = event.Records[0].s3.object.key

    // Download the JSON errors file from S3
    const getObjectResponse = await s3.send(
      new GetObjectCommand({ Bucket: bucket, Key: key })
    )

    // Add the contents of the JSON to the database
    if (getObjectResponse.Body) {
      /* example file path/key
         /uploads/organization_id/agency_id/reporting_period_id/expenditure_category_code/upload_id/{filename}
      */
      const result = JSON.parse(getObjectResponse.Body)

      // when the results array is empty then we know the file has passed validations
      let passed;
      if (result.length === 0) { passed = true } else { passed = false }

      const uploadId = await extractUploadIdFromKey(key)
      const input = { results: getObjectResponse.body, uploadId: uploadId, passed: passed}

[...]

PR #154, opened 2024-04-01
fix: remove expenditure-code from s3 url

as1729:
#139

As seen in the issue, this PR removes the expenditure category code from the S3 url for where the file is stored.

### DIFF
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -64,7 +64,7 @@ export function uploadWorkbook(
   uploadId: number,
   body: StreamingBlobPayloadInputTypes
 ) {
-  const folderName = `uploads/${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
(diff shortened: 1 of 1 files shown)


api/src/lib/aws.ts
```diff
@@ -64,7 +64,7 @@ export function uploadWorkbook(
   uploadId: number,
   body: StreamingBlobPayloadInputTypes
 ) {
-  const folderName = `uploads/${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.expenditureCategoryId}/${uploadId}/${upload.filename}`
+  const folderName = `uploads/${upload.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${uploadId}/${upload.filename}`
```
jakekreider, line 67: Related question for #135 - how stable do you expect these paths will be? Should I always find uploadId as the last node before filename?

jakekreider (approved):

api/src/lib/aws.ts
as1729: We expect this to be set in stone. I'm expecting the removal of the expenditure category code to be the last change needed here.

PR #155, opened 2024-04-03
cpfValidation handler

### LINKED ISSUE #135: Agency Report Submission - cpfValidation Lambda function

jakekreider:
Closes #135 

First pass at this - Any advice for testing the actual lambda terraform? Do I need to set it up in my own AWS account?

Thanks @as1729 for the assist in debugging!

### DIFF
--- a/api/src/functions/cpfValidation/cpfValidation.ts
+++ b/api/src/functions/cpfValidation/cpfValidation.ts
@@ -1,10 +1,15 @@
-import { https } from 'https'
-
+import {
+  S3Client,
+  GetObjectCommand,
+  DeleteObjectCommand,
+} from '@aws-sdk/client-s3'
+import { NodeJsClient } from '@smithy/types'
 import { S3Event, S3Handler } from 'aws-lambda'
 
+import { db } from 'src/lib/db'
 import { logger } from 'src/lib/logger'
 
-const apiEndpoint = 'https://example.com'
+const s3 = new S3Client({}) as NodeJsClient<S3Client>
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 export const handler: S3Handler = async (event: S3Event): Promise<void> => {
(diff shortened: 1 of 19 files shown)


api/src/functions/cpfValidation/cpfValidation.ts
```diff
@@ -1,28 +1,104 @@
...
+    const uploadId = extractUploadIdFromKey(key)
+    const input = {
+      results: result,
+      uploadId: uploadId,
+      passed: passed,
+    }
+    // There should be an existing validation Record in the DB that will need to be updated
+    const validationRecord = await db.uploadValidation.updateMany({
+      data: input,
+      where: { uploadId, passed: false },
```
as1729, line 74: one additional check here will be needed:
```js
where: { uploadId, passed: false, results: { equals: Prisma.JsonNull } }
```

jakekreider: Ah good call

as1729 (approved):

PR #158, opened 2024-04-04
fix: get user info from api

as1729:


as1729:
Addressed by #174 #175 #176 #177

Issue #132, opened 2024-03-20
Agency Report Submission - Database Changes


as1729:
### Why is this issue important?

See User Story here https://github.com/usdigitalresponse/cpf-reporter/issues/73

This is an issue in the sequence of tickets to ensure the CPF reporter can accept a user-submitted upload file and receive feedback on any errors that are present in the submitted file data.

### Blocks

- [x] #135

### Blocked By
None

### Current State

There are some tables available in the database that capture the upload validation data, but they are now outdated.

### Expected State

1. Ensure the latest database schema aligns with the Lucid Chart diagram here. Some of these changes may already exist in which case please verify it matches the ticket requirements.
2. Add a constraint in the `UploadValidation` table which ensures that for any `uploadId` there can only be one record where `results` is `null`.

<img width="604" alt="image" src="https://github.com/usdigitalresponse/cpf-reporter/assets/6497366/26035eac-fa56-4771-b582-dc1d6b15b0ba">
<img width="610" alt="image" src="https://github.com/usdigitalresponse/cpf-reporter/assets/6497366/8cbd05cc-c8c3-4d3c-ba05-b08d63038713">

### Implementation Plan
Note: we are not implementing `ValidationRules` table for this V1 April 10th deadline.
```
// schema.prisma

// Rename Upload to UploadFile and edit the fields
model UploadFile {
  id                    Int                  @id @default(autoincrement())
  filename              String
  uploadedById          Int
  uploadedBy            User                 @relation(fields: [uploadedById], references: [id])
  agencyId              Int
  agency                Agency               @relation(fields: [agencyId], references: [id])
  reportingPeriodId     Int
  reportingPeriod       ReportingPeriod      @relation(fields: [reportingPeriodId], references: [id])
  expenditureCategoryId Int?
  expenditureCategory   ExpenditureCategory? @relation(fields: [expenditureCategoryId], references: [id])
  createdAt             DateTime             @default(now()) @db.Timestamptz(6)
  updatedAt             DateTime             @default(now()) @db.Timestamptz(6)
  validations           UploadValidation[]
}

model UploadValidation {
  id                  Int            @id @default(autoincrement())
  uploadId            Int
  upload              Upload         @relation(fields: [uploadId], references: [id])
  passed                bool
  results                Json?         @db.JsonB
  initiatedById          Int
  initiatedBy            User                 @relation(fields: [initiatedById], references: [id])
[...]

PR #161, opened 2024-04-04
Feature/CPF-132 Remaining db changes

Vikariusu:
# Ticket #132 

## About
This ticket implements the remaining database changes:
- Removes the Subrecipient relation from the Upload model
- Removes the Organization relation from the Upload model
- Adjusts existing testing scenarios to match the new schema
- Per conversation with @as1729, we decided not to change Upload to UploadFile for now

### DIFF
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -108,8 +108,7 @@ export async function s3PutSignedUrl(
   const baseParams: PutObjectCommandInput = {
     Bucket: REPORTING_DATA_BUCKET_NAME,
     Key: key,
-    ContentType:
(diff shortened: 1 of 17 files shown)


as1729 (approved):

api/src/lib/aws.ts
```diff
@@ -108,8 +108,7 @@ export async function s3PutSignedUrl(
   const baseParams: PutObjectCommandInput = {
```
laurelkorwin, line 108: Sorry am very late to this since I was out last week but, do we need to update this path now that we don't have `organizationId` directly on `upload` anymore?

Issue #145, opened 2024-03-25
Agency Report Submission - CPF Validation - Prototype code cleanup


as1729:
### Why is this issue important?

See User Story here https://github.com/usdigitalresponse/cpf-reporter/issues/73

This is an issue in the sequence of tickets to ensure the CPF reporter can accept a user-submitted upload file and receive feedback on any errors that are present in the submitted file data.

### Current State

There exists `pydantic` schema and prototype code that can load an excel file from a local machine and generate any errors.

### Expected State

Take this existing schema and code and translate it into:
1. Easier to test modules/functions
2. Corresponding unit tests

### Implementation Plan

```py

from io import BytesIO

# Accepts a workbook from openpyxl
def validate_logic_sheet(logic_sheet: Any):
    errors = []
    try:
        # Cell B1 contains the version
        LogicSheetVersion(**{ "version": logic_sheet['B1'].value })
    except ValidationError as e:
        errors.append(f'Logic Sheet: Invalid {e}')
    return errors


def validate_cover_sheet(cover_sheet) -> Tuple[Optional[List[str]], Optional[Any]]:
    project_schema = None
    cover_header = get_headers(cover_sheet, 'A1:B1')
    cover_row = cover_sheet[2]
    row_dict = map_values_to_headers(cover_header, cover_row)
    try:
        CoverSheetRow(**row_dict)
    except ValidationError as e:
        errors.append(f"Cover Sheet: Invalid {e}")
        return (errors, None)

        # This does not need to be a silent failure. This would be a critical error.
    project_schema = SCHEMA_BY_PROJECT[row_dict['Project Use Code']]
    return (None, project_schema)

def validate_project_sheet(project_sheet) -> List[str]:
    errors = []
    project_headers = get_headers(project_sheet, 'C3:DS3')
    current_row = 12
    for project_row in project_sheet.iter_rows(min_row=13, min_col= 3, max_col=123, values_only=True):
        current_row += 1
        if is_empty_row(project_row):
            continue
        # print(project_row)
        row_dict = map_values_to_headers(project_headers, project_row)
        # print(row_dict['Capital_Asset_Ownership_Type__c'])
        try:
            project_schema(**row_dict)
        except ValidationError as e:
            errors.append(f"Project Sheet: Row Num {current_row} Error: {e}")
     return errors

def validate_subrecipient_sheet(subrecipient_sheet) -> List[str]:
    errors = []
[...]

PR #164, opened 2024-04-05
CPF validation refactor + tests

jakekreider:
Closes https://github.com/usdigitalresponse/cpf-reporter/issues/145

Note this is targeting another feature branch. This is mostly based off of the suggested refactor in the ticket, with a few fixes & more type hints. 

I didn't add exhaustive tests, but it should be pretty extensible - it takes a valid file and modifies directly for additional scenarios. Let me know if a different approach is preferred!


.gitattributes
```diff
@@ -19,3 +19,4 @@
 *.ttf               binary
 *.woff              binary
 *.woff2             binary
+*.xlsm              binary
```
jakekreider, line 22: This prevents git from treating the test data file as text & "helpfully" correcting line endings

as1729 (approved):

PR #176, opened 2024-04-10
Fix upload form errors

as1729:
#139

### DIFF
--- a/api/src/services/uploads/uploads.ts
+++ b/api/src/services/uploads/uploads.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '@prisma/client'
 import type {
   QueryResolvers,
   MutationResolvers,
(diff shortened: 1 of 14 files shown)


api/src/services/uploads/uploads.ts
```diff
@@ -20,8 +21,13 @@ export const upload: QueryResolvers['upload'] = ({ id }) => {
 export const createUpload: MutationResolvers['createUpload'] = async ({
   input,
 }) => {
+  const inputWithContext: Prisma.UploadUncheckedCreateInput = {
+    ...input,
+    uploadedById: context.currentUser.id,
+  }
+  console.log('inputWithContext', inputWithContext)
```
Vikariusu, line 28: nit: we should remove it once we know the upload flow is working

Issue #184, opened 2024-04-12
Agency Report Submission - Fix python validation errors


as1729:

ClaireValdivia:
4/15 convo - Aditya finished this and it seems resolved, but uncovered a new issue that Aditya will be creating that is still blocking being able to test upload validation

PR #186, opened 2024-04-12
fix: permissions for processValidationJSON function

as1729:
#184

Fixes the following `Access Denied` error. The lambda function should have the ability to `get` JSON files however it currently only has the ability to `put` JSON files. This PR fixes it.

### DIFF
--- a/terraform/functions.tf
+++ b/terraform/functions.tf
@@ -341,7 +341,7 @@ module "lambda_function-processValidationJson" {
 
   // Metadata
   function_name = "${var.namespace}-processValidationJson"
-  description   = "Reacts to S3 events and converts Excel files to JSON."
+  description   = "Reacts to S3 events and processes uploaded JSON files."
 
   // Networking
   attach_network_policy  = false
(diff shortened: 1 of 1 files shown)


terraform/functions.tf
```diff
@@ -357,23 +357,16 @@ module "lambda_function-processValidationJson" {
   policy_jsons                      = local.lambda_default_execution_policies
   attach_policy_statements          = true
   policy_statements = {
-    AllowDownloadExcelObjects = {
+    AllowDownloadAndDeleteJsonObjects = {
       effect = "Allow"
       actions = [
         "s3:GetObject",
         "s3:HeadObject",
+        "s3:DeleteObject",
```
TylerHendrickson, line 365: Was this intended to be `s3:PutObject`?

TylerHendrickson (changes requested):
@as1729 I'm a bit confused – the opening message for this PR says that this function needs
> the ability to `get` JSON files however it currently only has the ability to `put` JSON files.

But based on the code changes, it looks like:
- `s3:GetObject` permissions already exist
- `s3:DeleteObject` is being added (why?)
- There is no allowance for `s3:PutObject`

Am I missing something here?

as1729:
@TylerHendrickson I should clarify the intent here-

## Current State
Function has access to:
GET - XLSM files
PUT - JSON files

## Expected State
Function has access to:
GET       - JSON files
DELETE - JSON files

terraform/functions.tf
as1729: no, as mentioned in the latest comment. This function needs to only be able to `get` and `delete` the `.json` files

TylerHendrickson (approved):

PR #215, opened 2024-04-23
Update python errors schema

laurelkorwin:
- This adds more detail to validator error messages, including the tab, row and column as well as the message
- The tab and the row were straightforward to get already, column was a bit trickier. I ended up mapping the columns from the google doc from @as1729 straight to the Pydantic `Field` definitions. This means that we can access them by field in the event of an error by mapping the error's `loc` property to the field def, and accessing the `json_schema_extra` which stores the column
- To make this work consistently, I also specified in the model configs that we want to loc by field name and not by alias (see [here](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.loc_by_alias)

### DIFF
--- a/python/src/lib/workbook_validator.py
+++ b/python/src/lib/workbook_validator.py
@@ -9,6 +9,7 @@ from src.schemas.latest.schema import (
     LogicSheetVersion,
     SubrecipientRow,
     BaseProjectRow,
+    BaseModel
 )
 
 type Errors = List[WorkbookError]
@@ -47,6 +48,16 @@ def get_project_use_code(cover_sheet: Worksheet) -> str:
     cover_row = map(lambda cell: cell.value, cover_sheet[2])
     row_dict = map_values_to_headers(cover_header, cover_row)
     return row_dict["Project Use Code"]
+
+def get_erroring_column(SheetModelClass: BaseModel, e: ValidationError) -> str:
+    # For some reason getattr doesn't reliably work here, so using __fields__ as a dict instead
+    erroring_field = SheetModelClass.__fields__[e.errors()[0]['loc'][0]]
+    if (erroring_field and erroring_field.json_schema_extra):
+        erroring_column = erroring_field.json_schema_extra["column"]
+    else:
+        erroring_column = "Unknown"
+    return erroring_column
+
 
 def validate(workbook: IO[bytes]) -> Tuple[Errors, Optional[str]]:
     """Validates a given Excel workbook according to CPF validation rules.
@@ -117,11 +128,7 @@ def validate_cover_sheet(
     try:
         CoverSheetRow(**row_dict)
     except ValidationError as e:
-        erroring_field = CoverSheetRow.__fields__[e.errors()[0]['loc'][0]]
-        if (erroring_field and erroring_field.json_schema_extra):
-            erroring_column = erroring_field.json_schema_extra["column"]
-        else:
-            erroring_column = "Unknown"
+        erroring_column = get_erroring_column(CoverSheetRow, e)
         errors.append(WorkbookError(f"{COVER_SHEET} Sheet: Invalid {e}", f"{row_num}", f"{erroring_column}", COVER_SHEET))
         return (errors, None)
 
@@ -144,13 +151,7 @@ def validate_project_sheet(project_sheet: Worksheet, project_schema, project_use
         try:
             project_schema(**row_dict)
         except ValidationError as e:
-            # For some reason getattr doesn't reliably work here, so using __fields__ as a dict instead
-            #TODO figure out how to get the right row for the project use code
-            erroring_field = BaseProjectRow.__fields__[e.errors()[0]['loc'][0]]
-            if (erroring_field and erroring_field.json_schema_extra):
-                erroring_column = erroring_field.json_schema_extra["column"]
-            else:
-                erroring_column = "Unknown"
+            erroring_column = get_erroring_column(SCHEMA_BY_PROJECT[project_use_code], e)
             errors.append(WorkbookError(f"{PROJECT_SHEET} Sheet: Error {e}", f"{current_row}", f"{erroring_column}", PROJECT_SHEET))
     return errors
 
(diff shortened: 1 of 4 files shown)


laurelkorwin:
Column mappings have been triple checked and look good! This is ready for review / approval if folks feel good about it.

I'd recommend first minimizing `schema.py` since the changes are ginormous, and digesting the changes in `workbook_validator` and the associated test, and then looking over `schema.py`

python/src/lib/workbook_validator.py
```diff
@@ -121,7 +158,8 @@ def validate_project_sheet(project_sheet: Worksheet, project_schema) -> Errors:
         try:
             project_schema(**row_dict)
         except ValidationError as e:
-            errors.append(f"Project Sheet: Row Num {current_row} Error: {e}")
+            erroring_column = get_erroring_column(SCHEMA_BY_PROJECT[project_use_code], e)
```
as1729, line 161: Just tested the implementation and it looks good. One small adjustment I'd recommend is to handle capturing all the `errors()` that are captured. Here's one implementation you may want to consider for this which would replace the existing `get_erroring_column` function.

```py
from pydantic_core import ErrorDetails

"""
This function converts a list of ValidationError records for a single row into a list of WorkbookError records. Since each row has many different fields-- there can be multiple errors in a single row.
"""
def get_workbook_errors_for_row(SheetModelClass: BaseModel, errors: List[ErrorDetails], row_number: int, sheet_name: str) -> List[WorkbookError]:
    workbook_errors: List[WorkbookError] = []
    for e in errors:
        """
        Sample structure of the variable `e` here-
        https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.ErrorDetails
        {
         'type': 'string_type',
         'loc': ('Identification_Number__c',),
         'msg': 'Input should be a valid string',
         'input': None,
         'url': 'https://errors.pydantic.dev/2.6/v/string_type'
         }
        """
        field_name = e['loc'][0]
        workbook_column = SheetModelClass.__fields__[field_name].json_schema_extra['column']
        message = f'Error in field {field_name}-{e["msg"]}'
        workbook_errors.append(WorkbookError(message, row_number, workbook_column, sheet_name))
    return workbook_errors
    

try:
    project_schema(**row_dict)
except ValidationError as e:
    errors += get_workbook_errors_for_row(project_schema, e.errors(), current_row, PROJECT_SHEET)
        

```

python/src/lib/workbook_validator.py
```diff
@@ -94,20 +129,22 @@ def validate_cover_sheet(
...
+        errors.append(WorkbookError(f"{COVER_SHEET} Sheet: Invalid {e}", f"{row_num}", f"{erroring_column}", COVER_SHEET))
         return (errors, None)
 
         # This does not need to be a silent failure. This would be a critical error.
     project_schema = SCHEMA_BY_PROJECT[row_dict["Project Use Code"]]
     return (errors, project_schema)
 
 
-def validate_project_sheet(project_sheet: Worksheet, project_schema) -> Errors:
+def validate_project_sheet(project_sheet: Worksheet, project_schema, project_use_code) -> Errors:
```
as1729, line 147: I _think_ `project_use_code` may not be necessary here. I believe `SCHEMA_BY_PROJECT[project_use_code] == project_schema`. If that is not the case, feel free to ignore the comment.

python/src/lib/workbook_validator.py
```diff
@@ -8,9 +8,28 @@
...
+LOGIC_SHEET = "Logic"
+COVER_SHEET = "Cover"
+PROJECT_SHEET = "Project"
+SUBRECIPIENTS_SHEET = "Subrecipients"
+
+class WorkbookError:
+    message: str
+    row: str
+    col: str
+    tab: str
```
as1729, line 26: would recommend adding another field here to capture the `field_name`. We aren't using this yet in the front-end but I think it may be useful in the future to display it.

python/src/lib/workbook_validator.py
```diff
@@ -30,6 +49,22 @@ def get_project_use_code(cover_sheet: Worksheet) -> str:
     row_dict = map_values_to_headers(cover_header, cover_row)
     return row_dict["Project Use Code"]
 
+"""
+This method maps the thrown error to the impacted column in the spreadsheet.
+It does so by first getting the error's location (the loc property), which is the field name,
```
as1729, line 54: Thanks for figuring this out!

python/src/lib/workbook_validator.py
```diff
@@ -8,9 +8,28 @@
...
     LogicSheetVersion,
     SubrecipientRow,
+    BaseProjectRow,
+    BaseModel
 )
 
-type Errors = List[str]
+type Errors = List[WorkbookError]
+
+LOGIC_SHEET = "Logic"
```
as1729, line 17: Making these constants is great!

python/src/lib/workbook_validator.py
laurelkorwin: Oh hmm perhaps I misunderstood how the Pydantic / validator implementation was throwing things -- I had assumed based on the way the errors were being handled before that each cell would be thrown / handled separately and appended to the error list! I can def adjust to this though

python/src/lib/workbook_validator.py
laurelkorwin: Oooh I think you're right; I missed that since it was bubbled up from `validate_ cover_sheet`

python/src/lib/workbook_validator.py
as1729: I was a bit confused with this as well and wrote a small proof of concept code for my own understanding. It walks through how the spreadsheet rows formatted and passed into pydantic. Feel free to take a look here-
https://gist.github.com/as1729/d36bb2be853da5789afda032d794b070

My understanding after doing this is the following:
- Spreadsheet is converted into a list of row-dictionaries
- Each row-dictionary consists of field-header and field-value as key-value pairs.
- A pydantic schema class is initialized using the row-dictionary.
- Initialization of a pydantic-class validates all key-value pairs within the row-dictionary and raises a list of errors one for each field that is invalid

laurelkorwin: Gotcha, that makes sense! I've updated to handle all the errors in the list :)

as1729 (approved):

Issue #214, opened 2024-04-22
Agency Report Submission - Scope Uploads list by user-role


as1729:
## Current State
- All users regardless of role can see the same list of uploads

## Expected State
1. Staff users can only see their uploads in the list `user.id == uploads.uploadedById`
2. Agency Admin and USDR_Admin users can see any upload within their Organization-  `user.agency.organizationId == uploads.agency.organizationId`

## Components Affected
Uploads Table as seen below

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/6497366/2bffc2f8-a620-4a9b-99cd-55ecb36d553e)

## Implementation Plan
Point number 2 is already implemented in the `main` branch. Please verify
Point number 1 needs to be implemented.

PR #216, opened 2024-04-23
CPF-214 Scope uploads by user role

Vikariusu:
# Ticket #214 

## Overview

This PR implements the following functionality:
* Staff users can only see their own uploads (user.id == upload.uploadedById)
* Organization Admin and USDR Admin users can see any upload within their Organization (user.agency.organizationId == upload.agency.organizationId)
* Testing for all 3 roles (since organization admins and USDR admins have similar permissions, I added a reusable `uploadsBelongToOrganization()` function to follow DRY convention)

### DIFF
--- a/api/src/services/uploads/uploads.ts
+++ b/api/src/services/uploads/uploads.ts
@@ -7,28 +7,31 @@ import type {
 
 import { s3PutSignedUrl } from 'src/lib/aws'
 import aws from 'src/lib/aws'
+import { ROLES } from 'src/lib/constants'
 import { db } from 'src/lib/db'
 import { logger } from 'src/lib/logger'
 import { ValidationError } from 'src/lib/validation-error'
-import { ROLES } from 'src/lib/constants'
 
 export const uploads: QueryResolvers['uploads'] = () => {
-  const currentUser = context.currentUser;
+  const currentUser = context.currentUser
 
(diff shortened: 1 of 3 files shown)


api/src/services/uploads/uploads.ts
```diff
@@ -7,12 +7,32 @@ import type {
...
 import aws from 'src/lib/aws'
+import { ROLES } from 'src/lib/constants'
 import { db } from 'src/lib/db'
 import { logger } from 'src/lib/logger'
 import { ValidationError } from 'src/lib/validation-error'
 
 export const uploads: QueryResolvers['uploads'] = () => {
-  return db.upload.findMany()
+  const currentUser = context.currentUser
+
```
as1729, line 17: I'd like to make a slight clarification to the logic. The `organizationId` check should be done for all users regardless of whether they are staff. Only staff should have the added check of seeing only files they have uploaded. Here's what I'm thinking-

```js
const whereInputs = {
    agency: { organizationId: currentUser.agency.organizationId }
}

if (hasRole([ROLES.ORGANIZATION_STAFF])) {
    whereInputs.uploadedById = currentUser.id
}

return db.upload.findMany( { where: whereInputs } )
```

Vikariusu: @as1729 Adjusted the logic (and tests), thanks for the feedback!

as1729 (approved):

PR #226, opened 2024-04-26
Validate all sheets before returning

### LINKED ISSUE #221: 2024-05-13 Milestone - Version mismatch should not hard-return
## Current State
When there is an issue with the `Logic` sheet, the validator hard-returns.

## Expected State
When there is an issue with the `Logic` sheet, the validator continues to identify issues with other sheets.

### LINKED ISSUE #223: 2024-05-13 Milestone - Implement validation error severity
## Current State
`WorkbookError` class does not have a `severity` attribute.

## Expected State
`WorkbookError` class has a `severity` attribute that is of type Enum and defaults to `warn`. The options are:
- `err`
- `warn`
- `info`

All errors (except for version mismatch) are marked with `severity` of `err`.

## Implementation Plan
1. Update this line to send `WorkbookError(...., 'err')`. Any workbookerror generated by this function is an error. https://github.com/usdigitalresponse/cpf-reporter/blob/main/python/src/lib/workbook_validator.py#L84
2. Version mismatch only creates a warning `WorkbookError(..., 'warn')`. https://github.com/usdigitalresponse/cpf-reporter/blob/main/python/src/lib/workbook_validator.py#L141

jakekreider:
Fixes #221 

Small refactor to make the workbook more testable, then updated the logic to validate each sheet and return all errors at the end, instead of returning at the first one that fails. Let me know if that's not what was expected. 

_Update_
Closes #223 

Adds a `severity` property to `WorkbookError` class, defaults to `warn`.


python/src/lib/workbook_validator.py
```diff
@@ -97,38 +103,45 @@ def validate(workbook: IO[bytes]) -> Tuple[Errors, Optional[str]]:
...
+    errors += cover_errors
+
+    project_use_code = get_project_use_code(workbook[COVER_SHEET])
 
     """
-    5. Ensure all subrecipient rows are validated with the schema
+    3. Ensure all project rows are validated with the schema
     """
-    subrecipient_errors = validate_subrecipient_sheet(wb[SUBRECIPIENTS_SHEET])
+    errors += validate_project_sheet(workbook[PROJECT_SHEET], project_schema)
```
as1729, line 137: Note: we should only validate this if `project_schema` is not `None` and I believe there is a small chance it could be.

Open to solving it in either way of these two ways here:
1. add a conditional to only execute this line _if_ there is a `project_schema`
```py
if project_schema:
    validate_project_sheet
```
or within `validate_project_sheet` function
```py
if !project_schema:
    return [WorkbookError("Unknown Project"]
```

python/src/lib/workbook_validator.py
```diff
@@ -97,38 +103,45 @@ def validate(workbook: IO[bytes]) -> Tuple[Errors, Optional[str]]:
     See: https://openpyxl.readthedocs.io/en/stable/optimized.html
     If there is any trouble loading files from memory please see here: https://stackoverflow.com/questions/20635778/using-openpyxl-to-read-file-from-memory
     """
-    wb = load_workbook(filename=workbook, read_only=True)
+    try:
+        wb = load_workbook(filename=workbook, read_only=True)
+        return validate_workbook(wb)
```
as1729, line 108: I've noticed sometimes failures that aren't captured are showing up as "processing". It would be great if we can add an except block like so:
```py
from src.lib.logging import get_logger

logger = get_logger()
try:
    wb...
    return validate_workbook(wb, logger)
except Error as e:
    logger.error(f'Unexpected Validation Error: {e}')
    return [WorkbookError(f'Unable to validate workbook. Please reach out to grants-helpdesk@usdigitalresponse.org')]
finally:
    workbook.close()
```

jakekreider: Sounds good

python/src/lib/workbook_validator.py
jakekreider: Good call

jakekreider:
FYI @as1729 , I pulled #223 updates into this as well since it was touching related areas of the code.

as1729 (approved):

as1729:
Thank you!

PR #236, opened 2024-04-30
fix: ability to dump validation errors into json without type-errors

as1729:
#221

https://app.datadoghq.com/logs?query=service%3Acpf-reporter&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAY8vZ_rAuaptIAAAAAAAAAAYAAAAAEFZOHZaX3ZYQUFBbkloaFJWRHJ2clFBRQAAACQAAAAAMDE4ZjJmNjctZmFjMS00ODA1LThlZjQtOGRkZmJiNjczZmJi&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1714486414068&to_ts=1714487314068&live=true

Fixes the following error:
```
[ERROR] TypeError: Object of type ErrorLevel is not JSON serializable
Traceback (most recent call last):
  File "/opt/python/lib/python3.12/site-packages/datadog_lambda/wrapper.py", line 232, in __call__
    self.response = self.func(event, context, **kwargs)
  File "/var/task/src/lib/logging.py", line 54, in inner
    def inner(*args, **kwargs):
  File "/opt/python/lib/python3.12/site-packages/ddtrace/contrib/aws_lambda/patch.py", line 120, in __call__
    self.response = self.func(*args, **kwargs)
  File "/var/task/src/lib/logging.py", line 56, in inner
    return func(*args, **kwargs)
  File "/var/task/src/functions/validate_workbook.py", line 43, in handle
    save_validation_results(
  File "/var/task/src/functions/validate_workbook.py", line 108, in save_validation_results
    serialized = json.dumps(results)
  File "/var/lang/lib/python3.12/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/var/lang/lib/python3.12/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/var/lang/lib/python3.12/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
  File "/var/lang/lib/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
```

### DIFF
--- a/python/src/lib/workbook_validator.py
+++ b/python/src/lib/workbook_validator.py
@@ -1,13 +1,13 @@
 from enum import Enum
-from typing import IO, Any, Iterable, List, Optional, Tuple
+from typing import IO, Any, Iterable, List, Optional, Tuple, Union, Type, Dict
 
 from openpyxl import Workbook, load_workbook
 from openpyxl.worksheet.worksheet import Worksheet
 from pydantic import ValidationError
 from src.lib.logging import get_logger
 from src.schemas.latest.schema import (SCHEMA_BY_PROJECT, BaseModel,
-                                       CoverSheetRow, LogicSheetVersion,
-                                       SubrecipientRow)
+                                       CoverSheetRow, LogicSheetVersion, ProjectType,
+                                       SubrecipientRow, Project1ARow, Project1BRow, Project1CRow)
 
 type Errors = List[WorkbookError]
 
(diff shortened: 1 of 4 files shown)


python/src/lib/workbook_validator.py
```diff
@@ -34,16 +34,16 @@ class WorkbookError:
     col: str
     tab: str
     field_name: str
-    severity: ErrorLevel = ErrorLevel.WARN
+    severity: str = ErrorLevel.WARN.name
```
laurelkorwin, line 37: Happy to approve this as is but just curious, why not just implement a serializer method?

laurelkorwin (approved):

python/src/lib/workbook_validator.py
as1729: could you explain in code how that might look like? Happy to implement it! This was the first solution I was able to think of to fix the error I noticed so definitely open to other approaches.

laurelkorwin: This might not be the most elegant solution (I'm not the most up to date on Python best practices), but was thinking something like either subclassing `JSONEncoder` (see [here](https://docs.python.org/3/library/json.html#json.JSONEncoder)), or implementing a custom `toJSON` as here? https://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable?page=1&tab=scoredesc#tab-top on `ErrorLevel`. 

That said, I also think this approach is totally fine given it's a v small class / issue :)

PR #243, opened 2024-05-02
Chore: Upgrade actions/upload-artifact and actions/download-artifact

TylerHendrickson:
This PR upgrades older usages of `actions/upload-artifact` and `download-artifact` in GHA workflows to fix compatibility issues between newer and older versions.


TylerHendrickson:
Note: CI workflows (including for this PR) will fail until these changes are on `main`.

PR #253, opened 2024-05-07
Localstack configuration - WIP

jakekreider:
This plans and applies, but won't work (at a minimum, the localstack DB needs to be provided). Just getting feedback if this is an approach that will work. 

Should plan and apply locally with something like the following steps (from project root):
```
# bring up localstack
docker compose -f localstack/docker-compose.yml up -d 
# make sure it finishes setting up - docker logs cpf-reporter-localstack-main
# build redwoodjs app
yarn zip-it-and-ship-it api/dist/functions api/dist/zipballs
# build python wheel
cd python && poetry build && cd -
# plan and apply terraform
cd terraform && tflocal apply -var-file=local.tfvars -auto-approve; cd -
```

### DIFF
--- a/localstack/entrypoint/init-aws.sh
+++ b/localstack/entrypoint/init-aws.sh
@@ -1,7 +1,10 @@
 #! /bin/bash
+set -x
 
 export AWS_ACCESS_KEY_ID="test"
 export AWS_SECRET_ACCESS_KEY="test"
+
+apt install -y jq
 
 VALID_EMAILS=(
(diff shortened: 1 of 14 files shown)


TylerHendrickson (commented):
@jakekreider I left some questions as comments in a few places, but on the whole I think this is a pretty elegant approach that doesn't ultimately feel too hampered by the additional ternary conditions. This feels like the right approach to me.

PR #264, opened 2024-05-11
chore: adds logging around passage permissions failures

as1729:
#244

### DIFF
--- a/api/src/lib/auth.ts
+++ b/api/src/lib/auth.ts
@@ -56,6 +56,7 @@ export const getCurrentUser = async (
       const passageId = event.requestContext.authorizer?.claims?.sub
 
       if (!passageId) {
+        logger.error({ custom: event }, 'Passage ID not included.')
         throw new AuthenticationError('Passage ID not included.')
       }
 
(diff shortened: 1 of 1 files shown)


api/src/lib/auth.ts
```diff
@@ -65,7 +66,7 @@ export const getCurrentUser = async (
       })
 
       if (!user) {
-        logger.error(`User not found for passageId: ${passageId}`, event)
+        logger.error({ custom: event }, 'User not found.')
         return null
       }
```
laurelkorwin, line 72: Might also be worth adding a log message with the error in the `catch` block on line 95? Just on the off chance that the entire `try` is erroring?

laurelkorwin (approved):

api/src/lib/auth.ts
as1729: Ah! I missed this catch block. No wonder the `AuthenticationError` was not being logged in datadog. Makes a ton more sense now.

as1729: Addressed in a separate PR: https://github.com/usdigitalresponse/cpf-reporter/pull/265

PR #268, opened 2024-05-15
fix: ensure db is always a prismaclient

as1729:
https://github.com/usdigitalresponse/usdr-gost/issues/3052

### DIFF
--- a/api/src/lib/db.ts
+++ b/api/src/lib/db.ts
@@ -62,4 +62,11 @@ async function getPrismaClient() {
 /*
  * Instance of the Prisma Client
  */
-export let db: PrismaClient = getPrismaClient()
+getPrismaClient().then((db) => {
+  logger.info('Prisma Client initialized')
+  module.exports.db = db
+}).catch((error) => {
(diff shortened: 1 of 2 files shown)


api/src/lib/db.ts
```diff
@@ -47,16 +46,16 @@ async function createPrismaClient() {
...
   handlePrismaLogging({
     db,
     logger,
     logLevels: ['info', 'warn', 'error'],
   })
-})
+  return db
+}
+
+export const db: PrismaClient = await createPrismaClient()
```
TylerHendrickson, line 61: Back when we were first setting this up, I had issues implementing the top-level `await` in this manner. Fine with trying this out, but YMMV.

TylerHendrickson (approved):

api/src/lib/db.ts
as1729: ah missed this comment! Unfortunately it doesn't work with our commonjs setup so we need some other way to get this working.

TylerHendrickson (approved):

PR #269, opened 2024-05-17
Enable docker for local development

vshia:
As I was getting my local development environment to work with Docker, these are the changes I needed to make to be able to access `localhost:8910` from the host browser and for the seed code to work.

I also added some documentation to the README to get everything up an running in Docker.

### DIFF
--- a/redwood.toml
+++ b/redwood.toml
@@ -9,6 +9,7 @@
   bundler = "webpack"
   title = "Redwood App"
   port = 8910
+  host = "0.0.0.0"
   # You can customize graphql and dbauth urls individually too:
   # see https://redwoodjs.com/docs/app-configuration-redwood-toml#api-paths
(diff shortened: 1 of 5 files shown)


redwood.toml
```diff
@@ -9,6 +9,7 @@
   bundler = "webpack"
   title = "Redwood App"
   port = 8910
+  host = "0.0.0.0"
```
vshia, line 12: According to the [docs](https://redwoodjs.com/docs/app-configuration-redwood-toml#web), production uses `0.0.0.0` by default so this should not affect production. This was the main change to be able to hit the webserver from the host.

vshia: Actually, these [docs](https://redwoodjs.com/docs/app-configuration-redwood-toml#running-in-a-container-or-vm) say to do this and for the api. I'll update this.

as1729 (approved):

PR #274, opened 2024-05-17
Final changes to ReportingPeriods schema

laurelkorwin:


web/src/components/Organization/OrganizationPickListsCell/OrganizationPickListsCell.tsx
```diff
@@ -37,20 +33,6 @@ export const Success = ({
...
     <div>
-      <Label
-        name="reportingPeriodId"
-        className="rw-label"
-        errorClassName="rw-label rw-label-error"
-      >
-        Reporting Period
-      </Label>
-      <SelectField name="reportingPeriodId">
-        {organization.reportingPeriods.map((reportingPeriod) => (
```
as1729, line 48: can we update this to be a hidden form-field that has a value of `organization.preferences.current_reporting_period_id` rather than deleting this field?

Since this is critical for uploads this may cause issues if we remove it entirely.

laurelkorwin: Ah yes, will do!

laurelkorwin: Although, should I make `reportingPeriodId` nullable on `CreateUploadInput` until we're to the point where we're actually saving `preferences` on the organization?

laurelkorwin:
Closing in favor of #280

PR #275, opened 2024-05-18
Create global setup for Jest tests so db is present

jakekreider:
Since a lot of Redwood internals expect `api/src/lib/db` to always be an instantiated Prisma client (ref - https://github.com/redwoodjs/redwood/issues/3707), this creates Jest setup that runs before all tests to initialize it.

### DIFF
--- a/api/src/lib/db.ts
+++ b/api/src/lib/db.ts
@@ -64,6 +64,8 @@ export async function getPrismaClient() {
   return client
 }
 
+export default getPrismaClient
+
 /*
  * Instance of the Prisma Client
  */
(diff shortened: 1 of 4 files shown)


api/src/lib/db.ts
```diff
@@ -64,6 +64,8 @@ export async function getPrismaClient() {
   return client
 }
 
+export default getPrismaClient
```
jakekreider, line 67: I did this at first when attempting to use `globalSetup` (which won't work for tests), but left it in to simplify the setup code. Will remove if confusing.

as1729 (approved):

PR #276, opened 2024-05-18
add: update error messages and pydantic validators

vshia:
Addresses https://github.com/usdigitalresponse/usdr-gost/issues/3053

I'm not sure what the best way to go about this. Using `field_validators` is the pydantic approach but I didn't want to create a different field validator for every type of field.


as1729 (changes requested):
@vshia the solution looks good, the only thing I would add here is a test case that asserts that we receive the following error:
`Value is required for {info.field_name}`
- fee free to pick any field but as long as there's some coverage here we should be good

vshia:
@as1729 Added 3 tests. For some reason, if I set a project field to an empty string `" "` or whitespace `" "`, it fools the validation. I found this https://github.com/pydantic/pydantic/issues/2315#issuecomment-774510714, should we apply that  or something similar to prevent someone from adding a bunch of spaces?

as1729 (approved):

PR #286, opened 2024-05-24
Add localstack

vshia:
Communicates with localstack

The first time I run `yarn rw exec tests3`, it shows this in the localstack logs:
```
cpf-reporter-localstack-main  | 2024-05-24T19:23:27.437  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.CreateBucket => 200
cpf-reporter-localstack-main  | 2024-05-24T19:23:27.446  INFO --- [et.reactor-0] localstack.request.aws     : AWS s3.HeadObject => 404 (NoSuchKey)
```

The second time I run `yarn rw exec tests3`, it notes that the bucket was created (logs from in the redwood container).
```
$ yarn rw exec tests3
[STARTED] Generating Prisma client
[COMPLETED] Generating Prisma client
[STARTED] Running script
{"level":30,"time":1716578623250,"pid":598,"hostname":"56067b71b971","msg":"Initializing Prisma Client"}
------------ USING LOCALSTACK ------------
endpoint: http://localstack:4566
url http://localstack:4566/undefined/12345?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Credential=test%2F20240524%2Fus-west-2%2Fs3%2Faws4_request&X-Amz-Date=20240524T192343Z&X-Amz-Expires=60&X-Amz-Signature=6d10baae8818a7fc337aa749048c5ebf76cdf68150d654e4b41275bdb364ed42&X-Amz-SignedHeaders=host%3Bx-amz-server-side-encryption&x-id=PutObject
Failed to test s3.
BucketAlreadyOwnedByYou: Your previous request to create the named bucket succeeded and you already own it.
```

This tells me that the redwood docker container is communicating with the localstack container.


docker-compose.dev.yml
```diff
@@ -20,6 +20,8 @@ services:
       - SESSION_SECRET=super_secret_session_key_change_me_in_production_please
       - CI=
       - NODE_ENV=development
+      - LOCALSTACK_HOSTNAME=localstack
```
jakekreider, line 23: I also needed to add this env var to the `console` service to run the test script like `docker compose -f ./docker-compose.dev.yml run --rm -it console yarn rw exec tests3     ` - is that expected? Or should I have run this a different way?

scripts/tests3.ts
```diff
@@ -0,0 +1,48 @@
...
+    })
+    console.log("url", url)
+
+    const uploadParams: HeadObjectCommandInput = {
+      Bucket: REPORTING_DATA_BUCKET_NAME,
+      Key: key,
+    }
+    await s3.send(new CreateBucketCommand({
+      Bucket: REPORTING_DATA_BUCKET_NAME,
+    }))
```
jakekreider, line 40: Unless you were just wanting to use this script as a demo, maybe this could run first and generate a temp bucket, and then in a `finally` block issue a `DeleteBucketCommand` to clean up? That way it would succeed/fail deterministically?

jakekreider (commented):
Left a couple of comments but this your example works for me as expected!

docker-compose.dev.yml
vshia: ah you're right, will add

scripts/tests3.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    const response = await s3.send(new HeadObjectCommand(headParams))
+    console.log("--", response)
+
+    console.log("Delete file")
+    const deleteParams: DeleteObjectCommandInput = {
+      Bucket: REPORTING_DATA_BUCKET_NAME,
+      Key: key,
+    }
+    await s3.send(new DeleteObjectCommand(headParams))
```
jakekreider, line 71: Just realizing after my last comment - since this uses a prod bucket, maybe omit the delete after all. Or use a different bucket just for the test script. Sorry, bad idea on my part!

vshia: I added a `test` prefix and generated uuid for a suffix to make sure we don't mess with any real data.

jakekreider (approved):

PR #293, opened 2024-05-29
Adding versioning & new fields to python schema 

laurelkorwin:
- Adding versioning into the handling of our Pydantic schema
- Adding a new version & new fields
- Adding relevant unit tests
- Added documentation on how all this works (please comment liberally on this if I'm getting anything wrong / forgetting anything!)


docs/python-schema.md
```diff
@@ -0,0 +1,42 @@
+# Python Schema for Excel CPF Templates
```
as1729, line 1: Only note here is maybe we can link the developers to what the excel template looks like? In ARPA we commit to code all the versions of the input templates, and I'm wondering if something like that would make sense for CPF.

Alternatively we can link the Google Drive that includes the input templates.

as1729 (approved):
Thanks, the new structure seems more manageable in the long-term. Will look to test this out in staging.

docs/python-schema.md
laurelkorwin: Oooh yeah I think perhaps linking to the google drive would be nice / easiest to make sure they have access to different versions? I can add in a subsequent PR

PR #299, opened 2024-06-03
Add current_reporting_period_id to organization preferences where it doesn't exist

laurelkorwin:
* Add current reporting period ID to organization preferences where it doesn't exist
* Ran this migration locally to test -- it worked on my local DB and set the value correctly!

### DIFF
--- a/api/db/migrations/20240603182838_populate_organization_preferences/migration.sql
+++ b/api/db/migrations/20240603182838_populate_organization_preferences/migration.sql
@@ -1,0 +1,31 @@
+-- This is an empty migration.
+BEGIN;
+
+CREATE OR REPLACE FUNCTION populate_organization_preferences()
+RETURNS void AS $$
+DECLARE 
+    current_reporting_period_id INT;
+BEGIN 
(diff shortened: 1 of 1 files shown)


api/db/migrations/20240603182838_populate_organization_preferences/migration.sql
```diff
@@ -0,0 +1,30 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION populate_organization_preferences()
+RETURNS void AS $$
+DECLARE 
+    current_reporting_period_id INT;
+BEGIN 
+    -- Get one valid reporting period id --
+    SELECT id INTO current_reporting_period_id
```
tzinckgraf, line 9: Does this need to be ordered?

laurelkorwin: Good q! Not really -- this shouldn't touch production data, just run a one-off migration to fix data that's broken in staging -- so it's sufficient to assign it to any reporting period ATM

tzinckgraf (approved):

Issue #294, opened 2024-05-30
Subrecipient Validation - Ensure all projects have one valid subrecipient


as1729:

as1729:
@ClaireValdivia are you able to provide here what should go in each of the cells of this table when we identify a project that does not have a valid subrecipient?
Since this is a cross-sheet validation I wasn't sure what row/col/sheet-name to use. If you're able to add-in the exact error verbiage then this ticket should be ready for work.

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/6497366/58e5965d-4368-4964-8a63-bbee2b4488ea)

ClaireValdivia:
@as1729 just updated the definition of done section, lmk if you have any questions! referencing two columns is not our typical behavior but wondering if this feels feasible as I'm not sure of a better way to reference the error.

as1729:
That looks good to me 👍  thank you!

vshia:
Is there a difference between EIN and TIN? Is EIN just legacy (and if so, should we change it to TIN)?

ClaireValdivia:
@vshia EIN and TIN are referring to the same field, but unfortunately Treasury has named the field "EIN" in some places, so we need to be consistent with their naming conventions. 

For context, we don't want to get out of sync with naming of fields that are provided by Treasury and are included in the report we send to Treasury. The reason we were able to change a field name recently (Project Use Code > EC Code in the input template Cover sheet) is because that is a field name in our input template only and not provided by Treasury.

ClaireValdivia:
@vshia seeing a couple issues here for this [test upload](https://staging.cpf.usdr.dev/uploads/147). For this file, there are three project records 

- the first project record's validation is working as expected - I receive an error for the required field Subrecipient_TIN_c missing
- for the next two projects, however, they both do not have matching subrecipient UEI/TIN fields, so I expected two errors with the same message "You must submit a subrecipient record wiht the same UEI & TIN numbers entered for this project - one referencing tab "Project", Row 14, columns E, F, and the other referencing tab "Project", Row 15, columns E, F -- I received two errors with the correct Message and column reference, but they reference the tab "projects" instead of "Project," and both reference row 1 instead of 14 and 15

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/136839582/fc1383b8-c588-40d2-a4ba-fed16ec5233c)

vshia:
[...]

Issue #194, opened 2024-04-16
Agency Report Submission - Set Upload EC Code


as1729:
## Current State
When file is uploaded and validated, the EC Code on the Upload is `null` and appears as `Not Set` in the front-end.

## Expected State
When file is uploaded and validated, the EC Code is set.

## Implementation Plan
1. Update Python validation results to the following schema:
```
{
    errors: Array[string]
    projectUseCode: string 
}
```
In order to do this we will need to first find out how to derive the value of `projectUseCode`. To do this the following code will be helpful. This is a field in the "Cover" sheet of the Excel file that the python code is parsing.
```py
def get_project_use_code(cover_sheet: Worksheet) -> str:
    cover_header = get_headers(cover_sheet, "A1:B1")
    cover_row = map(lambda cell: cell.value, cover_sheet[2])
    row_dict = map_values_to_headers(cover_header, cover_row)
    return row_dict["Project Use Code"]
```
This will need to be added to the `workbook_validator.py`'s `validate` function as follows:
https://github.com/usdigitalresponse/cpf-reporter/pull/196/files#diff-b4bb27ac77a4277d0f1dd4db4e158e07418c6dff86bee251c51c1891e2c21ec9R28

Note: There will be typing changes as seen in the PR that need to be added as well.

2. Update `processValidationJSON` to accept the new input. Previous logic should now look at `errors` list to determine whether there were errors.
3. Update `processValidationJSON` to use the `projectUseCode` to set the EC Code value on the `Upload` object.
```js
    const expenditureCategory = await db.expenditureCategory.findFirst({
      where: { code: result.projectUseCode },
    })
    if (!expenditureCategory) {
      logger.error(
        `Expenditure category not found: ${result.projectUseCode} - key: ${key}`
      )
      throw new Error('Expenditure category not found')
    }

    // Update the Upload record with the expenditure category Id
    await db.upload.update({
      data: { expenditureCategoryId: expenditureCategory.id },
      where: { id: uploadId },
    })
```

PR #300, opened 2024-06-04
Add subrecipient validation

vshia:
Addresses https://github.com/usdigitalresponse/cpf-reporter/issues/294

Is there a good/easy way to test on the actual test files? I created test cases to verify this works but don't have a good way to test it on the actual files.

### DIFF
--- a/python/src/lib/workbook_validator.py
+++ b/python/src/lib/workbook_validator.py
@@ -219,12 +219,19 @@ def validate_workbook(workbook: Workbook) -> Tuple[Errors, Optional[str]]:
     3. Ensure all project rows are validated with the schema
     """
     if project_schema:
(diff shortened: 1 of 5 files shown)


TylerHendrickson:
@vshia re

> Is there a good/easy way to test on the actual test files? I created test cases to verify this works but don't have a good way to test it on the actual files.

It's a bit easy to miss, but if you look at the end of `python/src/lib/workbook_validator.py`, you'll see that it includes an `if __name__ == '__main__'` block that enables file validation from the CLI. If you have a test file you want to validate, you can run the validator against it like this (from within the `python/` directory):
```cli
poetry run python -m src.lib.workbook_validator /path/to/file_to_validate.xlsm
```
Note that #194 provides a link to where you can find and download workbook files for testing.

TylerHendrickson (approved):

Issue #305, opened 2024-06-06
[Issue]: Incorrect error when EC Code is blank


ClaireValdivia:
### Why is this issue important?

only one error is needed when EC Code is blank

### Current State

When EC Code is missing from a file, two errors are generating:

![Image](https://github.com/usdigitalresponse/cpf-reporter/assets/136839582/ee71fb67-05e6-44e1-87b1-717331c52c2e)

### Expected State

We only need the the error that says "Value is required for detailed_expenditure_category" in row 2, column B

Remove the error for row 2, column A - this is just help text for the sheet and there is no value that can be submitted

### Implementation Plan

The following functions need to be added...

ClaireValdivia:
This is looking good to me in staging!

vshia:
Here are the updates for the error text: https://github.com/usdigitalresponse/cpf-reporter/pull/307

ClaireValdivia:
@vshia The no projects message is looking good now!

However, when I try to test an upload with no EC Code selected in the Cover sheet, I get stuck on the Validation in Progress page and never get to validation results. ex upload in staging: https://staging.cpf.usdr.dev/uploads/146

vshia:
https://github.com/usdigitalresponse/cpf-reporter/pull/322 should address this now

ClaireValdivia:
this is looking good in staging now, thank you @vshia !

PR #307, opened 2024-06-06
Adjust error messages

vshia:
Due to a macro in the worksheet, the A2 field stays in sync with B2 so we don't need to show the invalid error. We could actually remove all validation on A2 if we want too but I left it in for now to be safe.

https://github.com/usdigitalresponse/cpf-reporter/issues/305


TylerHendrickson (approved):
@vshia Looks great – approved!

TylerHendrickson:
Just to add: the code updated in this PR doesn't seem like it should affect the API tests – hence the approval. However, whatever is causing the API unit tests to fail needs to be addressed (here or in another PR) before we're able to merge.

Issue #313, opened 2024-06-11
[Issue]: User's agency is not reflected in upload dropdown


ClaireValdivia:
### Why is this issue important?

If the user's agency is not the default selected, we expect to see files uploaded to the wrong agency frequently which will cause a lot of data issues

### Current State

in staging, when click Upload Workbook and are brought to the new workbook page, "Main Agency" is autoselected from the dropdown regardless of the users' agency

### Expected State

- On the new uploads page, organization staff users should only see their own agency in the dropdown, and it should be auto-selected as their own agency. Organization staff users should not be able to select any other agency from the dropdown.
- organization admin should be able to select any agency from the dropdown on the new upload page, but their own agency should be auto-selected when they go to the page

### Implementation Plan

The following functions need to be added...

ClaireValdivia:
Tested in staging, looks great!!

PR #335, opened 2024-06-16
CPF-313 Conditional agency dropdown

Vikariusu:
# Ticket #313 

## Changes
- Staff users
   - only see their own agency in the dropdown
   - their own agency is auto-selected
- Organization and USDR admins
    - see all agencies under their organization
     - their own agency is auto-selected

### Some small UI improvements 
Before
Screenshot 2024-06-16 at 3 36 31 PM

After
Screenshot 2024-06-16 at 3 36 02 PM

### DIFF
--- a/api/src/services/agencies/agencies.ts
+++ b/api/src/services/agencies/agencies.ts
@@ -35,7 +35,7 @@ export const deleteAgency: MutationResolvers['deleteAgency'] = ({ id }) => {
   })
 }
 
-export const agenciesAvailableForUpload = async () => { 
+export const agenciesAvailableForUpload = async () => {
   try {
     const agencies = []
     const user = await db.user.findFirst({
@@ -59,7 +59,9 @@ export const agenciesAvailableForUpload = async () => {
       agencies.push(agency)
       return agencies
     } else {
-      const allAgenciesUnderOrganization = agenciesByOrganization({ organizationId: organizationId })
+      const allAgenciesUnderOrganization = agenciesByOrganization({
+        organizationId: organizationId,
+      })
       return allAgenciesUnderOrganization
     }
   } catch (error) {
--- a/api/src/services/agencies/agencies.ts
+++ b/api/src/services/agencies/agencies.ts
@@ -1,5 +1,6 @@
 import type { QueryResolvers, MutationResolvers } from 'types/graphql'
 
+import { ROLES } from 'src/lib/constants'
 import { db } from 'src/lib/db'
 import { logger } from 'src/lib/logger'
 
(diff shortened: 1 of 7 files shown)


api/src/services/agencies/agencies.ts
```diff
@@ -35,6 +36,40 @@ export const deleteAgency: MutationResolvers['deleteAgency'] = ({ id }) => {
   })
 }
 
+export const agenciesAvailableForUpload = async () => {
```
Vikariusu, line 39: Named this resolver functionally, although it might be a good idea to investigate whether we can merge it with the existing non-role-specific `agenciesByOrganization` resolver.

web/src/components/Organization/OrganizationPickListsCell/OrganizationPickListsCell.tsx
```diff
@@ -6,19 +6,30 @@ import type {
 import { Label, SelectField, HiddenField } from '@redwoodjs/forms'
 import type { CellSuccessProps, CellFailureProps } from '@redwoodjs/web'
 
+import { useAuth } from 'src/auth'
+
+export const beforeQuery = () => {
+  // https://redwoodjs.com/docs/7.4/cells#beforequery
```
Vikariusu, line 12: Using `useAuth()` in `beforeQuery` gives the following eslint error: `React Hook "useAuth" is called in function "beforeQuery" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks`. Both `useAuth` and `beforeQuery` hooks are coming from Redwood, so renaming them is not really feasible.

api/src/services/agencies/agencies.ts
```diff
@@ -35,6 +36,40 @@ export const deleteAgency: MutationResolvers['deleteAgency'] = ({ id }) => {
...
+    const user = await db.user.findFirst({
+      where: { id: context.currentUser.id },
+    })
+
+    if (!user) {
+      logger.error('User not found')
+      return
+    }
+
+    const agency = await db.agency.findFirst({
```
as1729, line 51: I think we can simplify this a bit further since `currentUser` includes `agency` and we're already looking up the user from the DB when we retrieve `currentUser` - https://github.com/usdigitalresponse/cpf-reporter/blob/main/api/src/lib/auth.ts#L66

```js
async function agenciesAvailableForUpload() {
  if (currentUser.role === ..STAFF) {
      return [currentUser.agency]
  } else {
      return await agenciesByOrganization({ organizationId: currentUser.agency.organizationId })
  }
}
```

as1729 (approved):
Looks good minor refactor requested but can be done in a separate follow-up PR as the code-as-is is good for testing in staging.

api/src/services/agencies/agencies.ts
Vikariusu: Reopened the PR with this change here: https://github.com/usdigitalresponse/cpf-reporter/pull/337

PR #338, opened 2024-06-17
Generate treasury report for Projects 1A, 1B, 1C

vshia:
Ticket: https://github.com/usdigitalresponse/cpf-reporter/issues/315, https://github.com/usdigitalresponse/cpf-reporter/issues/316, https://github.com/usdigitalresponse/cpf-reporter/issues/317


python/src/schemas/schema_V2024_05_24.py
```diff
@@ -78,272 +81,515 @@ class CapitalAssetOwnershipType(str, Enum):
...
         )
     )
     Subrecipient_TIN__c: constr(strip_whitespace=True, min_length=9, max_length=9) = (
         Field(
             ...,
             serialization_alias="Subrecipient TIN",
-            json_schema_extra={"column": "F"},
+            json_schema_extra={
+                "column": "F",
+                "output_column_1A": "E",
```
as1729, line 179: can we name this `treasury_report_col_1A` `treasury_report_col_1B` and `treasury_report_col_1C`?

as1729:
@vshia it would be great if we can merge the changes in the schema `python/src/schemas/schema_V2024_05_24.py` as they look good. Are you able to open a separate PR with just those changes?

vshia:
@as1729 Unfortunately, I think it'll be hard to extract just the schema changes. There have been alot of lint changes which may result in alot of conflicts.

as1729:
@vshia one thing we would need here before we are ready to merge is the terraform changes to add the new lambda function.

You'll need to add a separate module called `lambda_function-<function_name>` similar to the pattern here: https://github.com/usdigitalresponse/cpf-reporter/blob/main/terraform/functions.tf#L445

The function will need to have permissions to retrieve the appropriate upload files from S3 as well as upload `.csv` files into s3. The pattern should be already available by looking at the module.

Let me know if you run into any issues adding that code.

laurelkorwin (approved):

python/src/functions/generate_treasury_report.py
```diff
@@ -0,0 +1,469 @@
...
+        project_use_code=project_use_code,
+    )
+
+    ### 2) Download the existing output XLSX file. If it doesn't exist, download
+    #    the template file.
+    output_file = tempfile.NamedTemporaryFile()
+    if project_agency_id_to_row_map:
+        download_file(
+            s3_client,
+            f"treasuryreports/{organization_id}/{current_reporting_period_id}/{user.id}",
```
as1729, line 121: @vshia one-small change here. Can we remove the `/{user.id}` prefix? The reports should be saved on the reporting period level and not be user-specific.

Apologies for not communicating this design change earlier. The change was made to ensure that the treasury report is the same regardless of who the user is for the entire organization.

as1729 (changes requested):
Small change requested to remove `{user.id}/` prefix wherever it exists as it is no longer needed.

as1729 (approved):

PR #340, opened 2024-06-18
tz/297/aws step functions

tzinckgraf:
Added AWS pieces for step functions.

There is a wrapper function in the `aws.ts` file.

I have added additional scripts and wrappers for testing using localstack.

The tests include a javascript and a python lambda function.
The javascript lambda function might be a bit off, but the python one works fine, so all the pipes work.


as1729 (approved):
Looks good @tzinckgraf ! I initially thought there were going to be changes to this PR but no changes required based on the updated treasury generation setup. The changes mostly affected the lambda functions directly.

PR #343, opened 2024-06-23
Provide DB client to validator function when creating user

### LINKED ISSUE #284: Bug - Unable to create users
## Current State

When attempting to create a user in staging the front-end receives a something went wrong error:

https://app.datadoghq.com/logs?query=service%3Acpf-reporter%20%40requestId%3Af1edd493-a1d1-4dee-8036-143e707f246f&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&sort=time&storage=hot&stream_sort=desc&view=spans&viz=stream&from_ts=1716562200000&to_ts=1716562800000&live=false

```
{
    "errors": [
        {
            "message": "Something went wrong.",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "createUser"
            ]
        }
    ],
    "data": null
}
```

## Expected State
Admin and USDR-admin user roles should be able to create and manage users from the Staging and production environments

## Implementation Steps
1. Verify if this is an issue only in the staging environment or does this happen locally also.
2. Figure out why there appears to be an issue connecting to the database only in this mutation and not other mutations.

## Stack trace
```
PrismaClientInitializationError: Authentication failed against database server at `cpfreporter-postgres-20231207000710979800000001.cluster-cwj9vaqylhtu.us-west-2.rds.amazonaws.com`, the provided database credentials for `postgres` are not valid.

Please make sure to provide valid database credentials for the database server at `cpfreporter-postgres-20231207000710979800000001.cluster-cwj9vaqylhtu.us-west-2.rds.amazonaws.com`.
    at r (/var/task/node_modules/@prisma/client/runtime/library.js:112:2493)
    at async xt.transaction (/var/task/node_modules/@prisma/client/runtime/library.js:111:11894)
    at async Proxy._transactionWithCallback (/var/task/node_modules/@prisma/client/runtime/library.js:127:9458)
    at async validateUniqueness (/var/task/node_modules/@redwoodjs/api/dist/validations/validations.js:408:10)
```

jakekreider:
Fixes #284 
I can't reproduce the error locally, but that's likely because the DB is created differently in local vs staging. But looking at the [validator source](https://github.com/redwoodjs/redwood/blob/a3879f37a717105893455028027fe10f2890e853/packages/api/src/validations/validations.ts#L720), if the db isn't provided, it uses the default, which wouldn't work for staging. I think this should fix it in staging, but I'm not sure how to test that.

### DIFF
--- a/api/src/services/users/users.ts
+++ b/api/src/services/users/users.ts
@@ -186,7 +186,7 @@ export const createUser: MutationResolvers['createUser'] = async ({
   return validateUniqueness(
     'user',
     { email },
-    { message: 'This email is already in use' },
+    { message: 'This email is already in use', db: db },
     (db) => db.user.create({ data: input })
   )
 }


api/src/services/users/users.ts
```diff
@@ -186,7 +186,7 @@ export const createUser: MutationResolvers['createUser'] = async ({
   return validateUniqueness(
     'user',
     { email },
-    { message: 'This email is already in use' },
+    { message: 'This email is already in use', db: db },
```
Vikariusu, line 189: Nice catch, this seems to be also mentioned in the [Redwood docs](https://docs.redwoodjs.com/docs/7.4/services#arguments)
"If you are overwriting the DATABASE_URL in your `src/lib/db` instantiation of the PrismaClient, you need to use this option. If not provided, a vanilla `new PrismaClient()` is used to run the callback that will not respect any custom configurations not represented in your `prisma.schema`"

Vikariusu (approved):

Issue #311, opened 2024-06-11
[Issue]: File with old version not generating results or displaying correct error message


ClaireValdivia:
### Why is this issue important?

Users should be able to upload a file with an old version number and see a warning message that lets them know the version is not the current one.

### Current State

When trying to upload [this file](https://drive.google.com/file/d/1e5_RDJODmopAZV9R2D-jewJLYOy6WU69/view?usp=drive_link), it gets stuck on the Validation in Progress screen and does not show the results. When this was previously working, the warning message was incorrect (unable to recreate currently due to the stuck in "validation in progress" issue)

### Expected State

If the workbook version is not the current version, the user should receive the following on the validation results page: 

Level: Warning
Message: “Upload template version ([uploaded template version]) is [older/ newer] than the latest input template ([latest template version])”
Tab: Logic
Row: 1
Col: B

### Implementation Plan

The following functions need to be added...

vshia:
https://github.com/usdigitalresponse/cpf-reporter/pull/342 should add the warning copy and fix the hanging validation

ClaireValdivia:
@vshia thanks so much!! I'm now able to get to the validation results page, but still seeing a couple issues:

- It looks like this file is not passing validation even though there is only a warning displaying. We would expect that if there are no Errors in the level column that validation would pass.
- Could we clean up the text of the message to read "“Upload template version ([uploaded template version]) is [older/ newer] than the latest input template ([latest template version])”? It looks like there's extra text before and after this in the Message.

![image](https://github.com/usdigitalresponse/cpf-reporter/assets/136839582/51b031c6-7a62-4155-bc74-f00538ad0caf)

ClaireValdivia:
@vshia this is looks great now, thank you!

PR #344, opened 2024-06-24
Address warning messages

vshia:
https://github.com/usdigitalresponse/cpf-reporter/issues/311

### DIFF
--- a/api/src/functions/processValidationJson/processValidationJson.ts
+++ b/api/src/functions/processValidationJson/processValidationJson.ts
@@ -117,7 +117,7 @@ export const processRecord = async (
     const result: ResultSchema = JSON.parse(strBody) || []
 
     // when the results array is empty then we know the file has passed validations
-    const passed = (result?.errors || []).filter((e) => e.severity != Severity.Error).length === 0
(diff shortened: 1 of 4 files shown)


api/src/functions/processValidationJson/processValidationJson.ts
```diff
@@ -105,7 +117,7 @@ export const processRecord = async (
     const result: ResultSchema = JSON.parse(strBody) || []
 
     // when the results array is empty then we know the file has passed validations
-    const passed = result.errors.length === 0
+    const passed = (result?.errors || []).filter((e) => e.severity != Severity.Error).length === 0
```
as1729, line 120: Should this be the following instead?

```js
// filter the error array to exclude warning and info severity and only include Severity.Error
// Ensure that the length of the resulting array is 0. If it is non-zero then there exists errors that are blocking validation
const passed = (result?.errors || []).filter((e) => e.severity == Severity.Error).length == 0
```

vshia: ah yes, you're right

as1729 (approved):

Issue #333, opened 2024-06-14
[Issue]: Clear out test files in prod [June 28]


ClaireValdivia:
### Why is this issue important?

We want partners to start the Q2 2024 reporting period out with a clean slate after testing core functionality of cpf reproter in june. 

### Current State

NA

### Expected State
 date - Friday June 28th
We will clear all files from the production environment for CPF Reporter and that going forward, partners will enter only real files for q2 reporting
Once the files are cleared, Aditya should email reporting-partners@usdigitalresponse.org to confirm 

### Implementation Plan

The following functions need to be added...

as1729:
```
Organization found: USDR
Deleting upload 1
{"level":30,"time":1719617249687,"pid":679,"hostname":"","upload_id":1,"msg":"deleted database record for upload"}
Deleting upload 2
{"level":30,"time":1719617249702,"pid":679,"hostname":"","upload_id":2,"msg":"deleted database record for upload"}
Deleting upload 9
{"level":30,"time":1719617249708,"pid":679,"hostname":"","upload_id":9,"msg":"deleted database record for upload"}
Deleting upload 10
{"level":30,"time":1719617249713,"pid":679,"hostname":"","upload_id":10,"msg":"deleted database record for upload"}
Deleting upload 11
{"level":30,"time":1719617249718,"pid":679,"hostname":"","upload_id":11,"msg":"deleted database record for upload"}
Deleting upload 12
{"level":30,"time":1719617249721,"pid":679,"hostname":"","upload_id":12,"msg":"deleted database record for upload"}
Deleting upload 13
{"level":30,"time":1719617249726,"pid":679,"hostname":"","upload_id":13,"msg":"deleted database record for upload"}
Deleting upload 14
{"level":30,"time":1719617249730,"pid":679,"hostname":"","upload_id":14,"msg":"deleted database record for upload"}
Deleting upload 15
{"level":30,"time":1719617249735,"pid":679,"hostname":"","upload_id":15,"msg":"deleted database record for upload"}
:: Script executed ::
[COMPLETED] Running script


Organization found: Wisconsin
Deleting upload 19
{"level":30,"time":1719617287523,"pid":855,"hostname":"","upload_id":19,"msg":"deleted database record for upload"}
:: Script executed ::
[COMPLETED] Running script


Organization found: Ohio
Deleting upload 3
[...]

PR #346, opened 2024-06-28
feat: add ability to delete and reset testfiles

as1729:
#333

This PR ensures that we can delete the test upload files that partners have added in production. Running the script does the following:
1. Identifies all uploads for the organization (optionally filtered by IDs if passed in)
2. Deletes the `UploadValidation` record
3. Deletes the `Upload` record
4. Deletes the file from s3

Unit tests have been added and ran the script locally.

### DIFF
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -1,4 +1,6 @@
 import {
+  DeleteObjectCommandInput,
+  DeleteObjectCommand,
   GetObjectCommand,
   HeadObjectCommand,
   HeadObjectCommandInput,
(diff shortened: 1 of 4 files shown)


api/src/lib/aws.ts
```diff
@@ -114,6 +116,20 @@ export async function s3PutSignedUrl(
   return url
 }
 
+export async function deleteUploadFile(upload: Upload) {
+  const fileKey = `uploads/${upload.agency.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.id}/${upload.filename}`
```
laurelkorwin, line 120: Any necessity for a simple try / catch here?

laurelkorwin (approved):

api/src/lib/aws.ts
as1729: good point, will add as a follow-on PR!

PR #347, opened 2024-06-28
fix: ensure terraform has the appropriate permissions

as1729:
#333

### DIFF
--- a/api/src/lib/aws.ts
+++ b/api/src/lib/aws.ts
@@ -119,12 +119,8 @@ export async function s3PutSignedUrl(
 }
 
 export async function deleteUploadFile(upload: Upload) {
-  try {
-    const fileKey = `uploads/${upload.agency.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.id}/${upload.filename}`
-    await s3DeleteObject(fileKey)
-  } catch (error) {
-    logger.error(error, 'Error deleting file from S3:')
-  }
+  const fileKey = `uploads/${upload.agency.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.id}/${upload.filename}`
+  await s3DeleteObject(fileKey)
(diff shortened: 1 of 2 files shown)


api/src/lib/aws.ts
```diff
@@ -117,8 +119,12 @@ export async function s3PutSignedUrl(
 }
 
 export async function deleteUploadFile(upload: Upload) {
-  const fileKey = `uploads/${upload.agency.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.id}/${upload.filename}`
-  await s3DeleteObject(fileKey)
+  try {
+    const fileKey = `uploads/${upload.agency.organizationId}/${upload.agencyId}/${upload.reportingPeriodId}/${upload.id}/${upload.filename}`
+    await s3DeleteObject(fileKey)
+  } catch (error) {
+    logger.error(error, 'Error deleting file from S3:')
```
TylerHendrickson, line 126: Minor nit: the `:` at the end of the log message is unnecessary since logs are structured and might not be displayed as `the log message: the error message` as a single string. I mention this since having it might make the log more difficult to query for in Datadog.

api/src/lib/aws.ts
```diff
@@ -117,8 +119,12 @@ export async function s3PutSignedUrl(
...
+  }
```
TylerHendrickson, line 127: I'm curious about why we're suppressing this error instead of letting it bubble up or re-throwing it after logging. I would think a hard error would make it more apparent that there was a permissions issue to address in the first place, but maybe I'm missing something.

as1729: that's a good point. let me remove the catch block for now. Since this function is only being executed by the script bubbling up the error would be ideal functionality

TylerHendrickson (approved):

laurelkorwin (approved):

PR #351, opened 2024-07-01
Lambda to prepare archive of CSV's

### LINKED ISSUE #318: Treasury Report - Lambda to zip all files and send email

jakekreider:
Closes #318 

Function to listen to SQS message on a new queue and generate an archive of CSV files for the given organization and reporting period. A few questions:

1. How will the email address of the user get to the next function #321 ? The plan in that one was to look at S3 files, but should that maybe be an SQS message as well? Or rolled all into this lambda?
2. Should the `report.zip` maybe be given some sort of timestamp in its name?
3. Some options for testing instructions:
	1. There's a shellscript that should run on its own, or can be used as a guide. Assumes presence of [`awslocal`](https://github.com/localstack/awscli-local), which is a localstack wrapper of the AWS CLI. 
	2. There's also a bats file - if there's interest in using that kind of framework, I can include the other bats submodules and flesh it out a bit more. If not, I can delete it.
	3. I can also delete both and just put the example here in the PR.

### DIFF
--- a/python/src/functions/create_archive.py
+++ b/python/src/functions/create_archive.py
@@ -9,7 +9,7 @@ from aws_lambda_typing.events import SQSEvent
 from mypy_boto3_s3.client import S3Client
 from src.lib.logging import get_logger, reset_contextvars
 
-S3_BUCKET = os.environ["S3_BUCKET"]
+S3_BUCKET = os.environ.get("S3_BUCKET", "test_bucket")
 
 
 @reset_contextvars
@@ -70,6 +70,7 @@ def create_archive(
             logger.info(f"Created archive file: {file.name}")
             upload_key = f"{target_key}report.zip"
             # upload the zip file to the target key
+
             s3_client.upload_file(file.name, S3_BUCKET, upload_key)
             logger.info(f"Uploaded archive file to {upload_key}")
             logger.info(f"Uploaded archive file to {upload_key}")
--- a/python/src/functions/create_archive.py
+++ b/python/src/functions/create_archive.py
@@ -1,0 +1,75 @@
+import json
+import os
+import tempfile
+import zipfile
+
+import boto3
+from aws_lambda_typing.context import Context
+from aws_lambda_typing.events import SQSEvent
+from mypy_boto3_s3.client import S3Client
+from src.lib.logging import get_logger, reset_contextvars
+
+S3_BUCKET = os.environ["S3_BUCKET"]
+
+
+@reset_contextvars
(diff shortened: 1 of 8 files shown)


python/src/functions/create_archive.py
```diff
@@ -0,0 +1,76 @@
...
+from aws_lambda_typing.context import Context
+from aws_lambda_typing.events import SQSEvent
+from mypy_boto3_s3.client import S3Client
+from src.lib.logging import get_logger, reset_contextvars
+
+S3_BUCKET = os.environ.get("S3_BUCKET", "test_bucket")
+
+
+@reset_contextvars
+def handle(event: SQSEvent, _context: Context):
```
as1729, line 16: As discussed on the Friday call the main change here is to update the input event to be a step-function event rather than an `SQSEvent`.

python/src/functions/create_archive.py
```diff
@@ -0,0 +1,76 @@
...
+    with tempfile.NamedTemporaryFile() as file:
+        with zipfile.ZipFile(file, "w") as zipf:
+            for target_file in target_files:
+                obj = s3_client.get_object(Bucket=S3_BUCKET, Key=target_file)
+                zipf.writestr(target_file, obj["Body"].read())
+
+            zipf.close()
+            file.flush()
+            logger.info(f"Created archive file: {file.name}")
+            upload_key = f"{target_key}report.zip"
```
as1729, line 71: this upload key of `report.zip` is fine because this value is what we will use to delete the zip-file whenever a new-generation has begun. There will only ever be one up-to-date .zip report.

as1729 (changes requested):
@jakekreider saw your message that you'll need to scale back your time over the next few weeks. I have put the changes-requested so we know what needs to be changed in the PR for the next person. Will look to push these changes out when I return to office, thank you!

terraform/functions.tf
```diff
@@ -515,3 +515,112 @@ module "lambda_function-cpfValidation" {
...
+  attach_policy_statements          = true
+  policy_statements = {
+    AllowDownloadExcelObjects = {
+      effect = "Allow"
+      actions = [
+        "s3:GetObject",
+        "s3:HeadObject",
+      ]
+      resources = [
+        # Path: treasuryreports/{organization_id}/{reporting_period_id}/{filename}.csv
```
vshia, line 549: In https://github.com/usdigitalresponse/cpf-reporter/issues/315, the path also includes a `{user.id}` field. @as1729 which path should be used?

PR #352, opened 2024-07-08
tz/298/uploads-by-expenditure-category

tzinckgraf:
Issue 298

Added function for uploads by expenditure category. Added test caes as well.


api/src/lib/aws.test.ts
tzinckgraf, line 1: Not sure why this whole file is showing up as new.

api/src/lib/aws.ts
```diff
@@ -97,13 +97,21 @@ async function sendHeadObjectToS3Bucket(bucketName: string, key: string) {
   await s3.send(new HeadObjectCommand(uploadParams))
 }
 
+export function getS3UploadFileKey(organizationId: number, upload: Upload | CreateUploadInput, uploadId?: number) {
+  if ('id' in upload) {
```
laurelkorwin, line 101: Could we add a small explanatory comment here for when we'd expect this to hit vs the `uploadId` we pass in?

api/src/services/uploads/uploads.test.ts
```diff
@@ -203,3 +211,45 @@ describe('downloads', () => {
     )
   })
 })
+
+describe('getUploads', () => {
+  scenario('returns most recent upload for one category',
+    async (scenario: UploadScenario) => {
+      mockCurrentUser(scenario.user.one)
+      await updateOrganization({
```
laurelkorwin, line 219: Could we pull out these `await` calls here and in the scenario below into a `beforeEach` block?

api/src/services/uploads/uploads.ts
```diff
@@ -147,3 +147,114 @@ export const Upload: UploadRelationResolvers = {
...
+      agency: { organizationId: organization.id },
+      id: { in: validUploadIds.map((v) => v.uploadId) },
+    },
+    include: { expenditureCategory: true },
+  })
+
+  const uploadsByExpenditureCategory = {}
+
+  // Get the most recent upload for each expenditure category and agency and set the S3 Object key
+  for (const upload of uploadsForPeriod)  {
```
laurelkorwin, line 225: Ultimately up to you and totally fine in a follow-up, but wondering if a small wrapper class for `uploadsByExpenditureCategory` with some getters and setters might make it easier for future devs to understand / edit this code?

laurelkorwin (approved):

api/src/services/uploads/uploads.ts
tzinckgraf: Good idea, but let's do this in a follow up PR. I think this PR is a blocker on some other work.

laurelkorwin (approved):

PR #355, opened 2024-07-08
Make subrecipients into JSON explicitly

laurelkorwin:
* Mapping each subrecipient to a JSON serializable object with `.model_dumps` so we can `json.dumps()` it in `validate_workbook.py`

### DIFF
--- a/python/src/lib/workbook_validator.py
+++ b/python/src/lib/workbook_validator.py
@@ -1,4 +1,5 @@
 from enum import Enum
+import json
 from typing import IO, Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 from openpyxl import Workbook, load_workbook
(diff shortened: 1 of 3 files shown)


python/src/schemas/schema_V2024_04_01.py
```diff
@@ -16,7 +16,7 @@
 from src.schemas.project_types import NAME_BY_PROJECT
 
 
-class StateAbbreviation(Enum):
+class StateAbbreviation(str, Enum):
```
laurelkorwin, line 19: Subclassing `str` for an Enum is a quick fix to make it JSON serializable

python/src/lib/workbook_validator.py
```diff
@@ -283,6 +284,9 @@ def validate_workbook(workbook: Workbook) -> Tuple[Errors, Optional[str]]:
             projects, subrecipients, version_string
         )
 
+    
+    subrecipients = [subrecipient.model_dump() for subrecipient in subrecipients]
```
laurelkorwin, line 288: I think this is the "supported" way in the latest version of Pydantic, [see docs](https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump)

I tested by hardcoding the same `results` object we use in `validate_workbook` in `workbook_validator` and running unit tests, then inspecting the output. `json.dumps` works and subrecipient output looks good 
Screenshot 2024-07-08 at 8 26 06 PM

as1729 (approved):

PR #363, opened 2024-07-12
Expose version string on return payload for validator

laurelkorwin:
* Expose version string on return payload from `validate_workbook` so that we can access it in `processValidationJson`

### DIFF
--- a/api/src/functions/processValidationJson/processValidationJson.test.ts
+++ b/api/src/functions/processValidationJson/processValidationJson.test.ts
@@ -53,6 +53,7 @@ describe('cpfValidation function', () => {
       },
     })
     organizationId = organization.id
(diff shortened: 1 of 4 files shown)


as1729:
@laurelkorwin just want to confirm - how does the version string help us in process JSON? Is it so we can store it with the subrecipient raw data?

laurelkorwin:
@as1729 for some reason I'm not able to reply to your comment directly but yes -- we're saving it with each `SubrecipientUpload` [here](https://github.com/usdigitalresponse/cpf-reporter/blob/48862912590a74cdaae4cc077387f4dbbcef0356/api/src/functions/processValidationJson/processValidationJson.ts#L251)

PR #366, opened 2024-07-16
Adding skeleton lambda for subrecipient file gen

laurelkorwin:
Adding terraform and a skeleton python file for subrecipient file gen lambda


TylerHendrickson (changes requested):
Requesting two trivial changes, but everything of substance looks good 👍 

PS - It occurs to me that this should be (but currently isn't) configured in our workspace settings. If you feel like it, adding the following entry to `.vscode/settings.json` will configure VSCode to automatically add a blank line to the end of a file (if one is missing) upon save:
```json
    "files.insertFinalNewline": true,
```

No worries if you'd prefer to not expand this PR – I can add it separately if you prefer.

as1729 (approved):

TylerHendrickson (approved):

Issue #296, opened 2024-05-30
Treasury Report - Graphql changes


as1729:
## Blocks
- [ ] https://github.com/usdigitalresponse/cpf-reporter/issues/296

## Blocked By
- [ ] https://github.com/usdigitalresponse/cpf-reporter/issues/297
- [ ] https://github.com/usdigitalresponse/cpf-reporter/issues/298

## Current State
does not exist

## Expected State
A `sendTreasuryReport` graphql mutation that compiles the data needed to kick-off the Amazon step function to generate and send the treasury report and starts the execution.

## Implementation Details
1. Add `sendTreasuryReport` mutation handler in the `users.ts` service:
```js
import { getUploadsByExpenditureCategory } from 'src/services/uploads/uploads'

export const sendTreasuryReport: MutationResolvers['sendTreasuryReport'] =
  async () => {
    const organization = await db.organization.findFirst({
      where: { id: context.currentUser.agency.organizationId },
    })
    const reportingPeriod = await db.reportingPeriod.findFirst({
      where: { id: organization.preferences.current_reporting_period_id },
    })
    const uploadsByExpenditureCategory = await getUploadsByExpenditureCategory()

    try {
      // Send the Treasury Report
      // For now, we just log that we're sending the report
      // Call amazon step function with the following parameters:
      /*
      {
        "reportingPeriod": reportingPeriod.name,
        "organization": organization.name,
        "email": context.currentUser.email,
        "uploadsByExpenditureCategory": uploadsByExpenditureCategory
      }
      */
      logger.info(uploadsByExpenditureCategory)
      logger.info('Sending Treasury Report')
      startStepFunctionExecution(
        organization,
        context.currentUser.email,
        uploadsByExpenditureCategory
      )
      return true
    } catch (error) {
      logger.error(error, 'Error sending Treasury Report')
      return false
    }
  }
```
3. Define the mutation correctly on the graphql side `users.sdl.ts`
```js
  type Mutation {
    sendTreasuryReport: Boolean! @requireAuth
  ...
  }
```

PR #369, opened 2024-07-23
CPF-296 Treasury Report GraphQL Mutation

Vikariusu:
# Ticket #296 

## Changes
* Added `sendTreasuryReport` GraphQL mutation
* `sendTreasuryReport` triggers aws step function
* TREASURY_STEP_FUNCTION_ARN` env variable is now required to successfully run sendTreasuryReport()


api/src/services/uploads/uploads.ts
```diff
@@ -258,3 +258,42 @@ export const getUploadsByExpenditureCategory = async () => {
...
+      const input = {
+        reportingPeriod: reportingPeriod.name,
+        organization: organization.name,
+        email: context.currentUser.email,
+        uploadsByExpenditureCategory,
+      }
+
+      await aws.startStepFunctionExecution(
+        arn,
+        undefined,
```
Vikariusu, line 290: As per [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sfn/command/StartExecutionCommand/): "Optional name of the execution. This name must be unique for your Amazon Web Services account, Region, and state machine for 90 days. For more information, see [Limits Related to State Machine Executions ](https://docs.aws.amazon.com/step-functions/latest/dg/limits.html#service-limits-state-machine-executions)
 in the Step Functions Developer Guide.

If you don't provide a name for the execution, Step Functions automatically generates a universally unique identifier (UUID) as the execution name."

Let me know if we would prefer to name this execution explicitly.

api/src/services/uploads/uploads.ts
```diff
@@ -258,3 +258,42 @@ export const getUploadsByExpenditureCategory = async () => {
 
   return uploadsByExpenditureCategory
 }
+
+export const sendTreasuryReport: MutationResolvers['sendTreasuryReport'] =
```
as1729, line 262: @Vikariusu can we add unit tests for this function? I wasn't sure if the comment below that you added is blocking your ability to do so.

Vikariusu: Just added testing

api/src/services/uploads/uploads.ts
as1729: I think its fine to leave this alone for now and use the UUID default. 👍

as1729 (approved):

Issue #348, opened 2024-06-28
Treasury Report - Unblock testing of file-generation in staging environment for USDR_ADMIN users


as1729:
## Context
While some of the initial file generation pieces are being merged into staging, there is no way to easily test this feature piece-meal via the staging environment. The objective of this ticket is to be able to create a way for a `USDR_ADMIN` user to directly call each of the new lambda functions that are being added download the resulting files being generated.

## Current State
N/A

## Expected State

`USDR_ADMIN` users see the following buttons at their disposal on the homepage. Note: these buttons should not appear for any other user-role and will be discarded once the Treasury Report feature is fully built-out.

Form Field 1 - Accepts a long text containing payload from the user.
Button 1 - Kick-off file generation
Button 2 - Download Project 1A file
Button 3 - Download Project 1B file
Button 4 - Download Project 1C file
Button 5 - Download Subrecipient file

All the "Download ***" buttons should create a GET signed-url for the s3 object containing the project/subrecipient file for the current reporting period of the organization.

The "Kick-off file generation" button should send any payload that the user has added in "Form Field 1" to all the lambda functions.

## Implementation Details
- Add a Redwood Form
- All download buttons can follow the same pattern as the existing Upload file download button except the s3 key will be different.
- Add documentation for users on how to fill out the form and structure the payload

ClaireValdivia:
Note for context - 295 & 296 & 320 are permanent solution for file generation. This issue (348) is a workaround to allow generation of files before those are done

PR #370, opened 2024-07-25
feat: FE and graphql scaffolding for manual treasury report generation

as1729:
There are 3 pieces to ensuring we are able to satisfy requirements of issue #348 

1. Front-end and GraphQL changes that accept payload from the user and parse it on the backend to submit to AWS.
2. Terraform permissions updates so the GraphQL lambda can invoke the treasury file generation lambdas.
3. GraphQL changes to `aws.ts` to support lambda invocation from the Redwoodjs lambda function.

This PR addresses point number 1.

Adds the following page that is only visible for USDR_ADMIN users.
image

## Unit tests
- There aren't any yet since the backend isn't really doing anything at the moment. These will come with the PR for # 3.


api/src/services/organizations/organizations.ts
```diff
@@ -74,6 +76,56 @@ export const getOrCreateOrganization = async (orgName, reportingPeriodName) => {
   }
 }
 
+export const downloadTreasuryFile: MutationResolvers['downloadTreasuryFile'] =
```
laurelkorwin, line 79: Would it be possible to add some unit tests for these, maybe in a followup?

api/src/services/organizations/organizations.ts
```diff
@@ -74,6 +76,56 @@ export const getOrCreateOrganization = async (orgName, reportingPeriodName) => {
...
+  async ({ input }) => {
+    const { fileType } = input
+    const { organizationId } = context.currentUser.agency
+
+    // get the organization
+    const organization = await db.organization.findUnique({
```
laurelkorwin, line 85: Do we want to throw an error here if organization is not found similar to the below?

web/src/components/TreasuryGeneration/DownloadTreasuryFiles/DownloadTreasuryFiles.tsx
```diff
@@ -0,0 +1,53 @@
...
+      console.log(fileLink)
+      console.log('Opening treasury file link in new tab..')
+      window.open(fileLink, '_blank').focus()
+    },
+    onError: (error) => {
+      toast.error(error.message)
+    },
+  })
+  const onClick = (event) => {
+    console.log(event.target.getAttribute('name'))
```
laurelkorwin, line 27: Nit: remove?

web/src/components/TreasuryGeneration/DownloadTreasuryFiles/DownloadTreasuryFiles.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+`
+
+const DownloadTreasuryFiles = () => {
+  const [downloadTreasuryFile] = useMutation(DOWNLOAD_TREASURY_FILE, {
+    onCompleted: ({ downloadTreasuryFile }) => {
+      const { fileLink } = downloadTreasuryFile
+      console.log(fileLink)
+      console.log('Opening treasury file link in new tab..')
```
laurelkorwin, line 19: Nit: remove?

web/src/components/TreasuryGeneration/NewTreasuryGenerationForm/NewTreasuryGenerationForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const { payload } = data
+
+    onSave({
+      payload,
+    })
+  }
+
+  return (
+    <Form
+      onSubmit={onSubmit}
```
laurelkorwin, line 39: Nit: Would it be simpler to just do `() => onSave({ payload })` here rather than creating a new func above?

laurelkorwin (approved):

api/src/services/organizations/organizations.ts
as1729: good call - done!

api/src/services/organizations/organizations.ts
as1729: yes! good call - done 👍

as1729 (commented):
@laurelkorwin one more set of unit tests coming up shortly for the `kickOff**` function, but otherwise all the comments should be addressed.

as1729:
@laurelkorwin ready for a re-review. I have updated this to remove all the graphql.d.ts type files. Hopefully by doing this they are once and for all gone from the repository. It is already untracked in `.gitignore` but when I made that PR I didn't permanently remove the files which is why it keeps popping up again.

laurelkorwin (approved):

PR #371, opened 2024-07-25
chore: ensure graphql lambda function has the correct IAM permissions

as1729:
There are 3 pieces to ensuring we are able to satisfy requirements of issue #348 

1. Front-end and GraphQL changes that accept payload from the user and parse it on the backend to submit to AWS.
2. Terraform permissions updates so the GraphQL lambda can invoke the treasury file generation lambdas.
3. GraphQL changes to `aws.ts` to support lambda invocation from the Redwoodjs lambda function.

This PR addresses point number 2.


terraform/functions.tf
```diff
@@ -683,5 +694,150 @@ module "lambda_function-treasuryReportGeneration" {
...
   }
 }
+
+
+module "step_function" {
+  source = "terraform-aws-modules/step-functions/aws"
+
+  name       = "generate-treasury-report"
+  definition = <<EOF
+{
```
as1729, line 710: Do not use this since this causes issues with the linter locally.

Provide a HCL terraform object and wrap it in jsonencode(``)

as1729: another option is to define the definition as a terraform local and pass it in as a one-liner. and use the local to define as an output.

terraform/treasury_generation_lambda_functions.tf
```diff
@@ -0,0 +1,174 @@
...
+    StepFunctionTrigger = {
+      principal  = "states.amazonaws.com"
+      source_arn = module.treasury_generation_step_function.state_machine_arn
+    }
+
+  }
+}
+
+module "lambda_function-treasuryReportGeneration" {
+  for_each = toset(["1A", "1B", "1C"])
```
TylerHendrickson, line 77: Given the extreme similarity between each of the Lambda functions generated by this module definition, I think it would be better to eliminate the `for_each` and define a single Lambda function for report generation. 

Unless there's a clear operational benefit for having distinct functions (or clear detriment to having them combined), I think merging is justified given that each Lambda in the `for_each` shares the following criteria:
1. Invocation source(s)
2. Execution policy
3. Handler callable (NB: inspecting a payload value feels like a more appropriate way to guide 1A/B/C selection than an env var)

terraform/treasury_generation_step_function.tf
```diff
@@ -0,0 +1,137 @@
...
+          {
+            "StartAt" : "Generate Project 1A File",
+            "States" : {
+              "Generate Project 1A File" : {
+                "Type" : "Task",
+                "Resource" : "arn:aws:states:::lambda:invoke",
+                "OutputPath" : "$.Payload",
+                "Parameters" : {
+                  "Payload.$" : "$",
+                  "FunctionName" : module.lambda_function-treasuryReportGeneration["1A"].lambda_function_arn
```
TylerHendrickson, line 22: See my previous comment; I think it would be better to inject a "report type" parameter into the payload and have `FunctionName` point to `module.lambda_function-treasuryReportGeneration.lambda_function_arn`.

terraform/treasury_generation_step_function.tf
```diff
@@ -0,0 +1,137 @@
...
+          {
+            "StartAt" : "Generate Project 1B File",
+            "States" : {
+              "Generate Project 1B File" : {
+                "Type" : "Task",
+                "Resource" : "arn:aws:states:::lambda:invoke",
+                "OutputPath" : "$.Payload",
+                "Parameters" : {
+                  "Payload.$" : "$",
+                  "FunctionName" : module.lambda_function-treasuryReportGeneration["1B"].lambda_function_arn
```
TylerHendrickson, line 50: (see above)

terraform/treasury_generation_step_function.tf
```diff
@@ -0,0 +1,137 @@
...
+          {
+            "StartAt" : "Generate Project 1C File",
+            "States" : {
+              "Generate Project 1C File" : {
+                "Type" : "Task",
+                "Resource" : "arn:aws:states:::lambda:invoke",
+                "OutputPath" : "$.Payload",
+                "Parameters" : {
+                  "Payload.$" : "$",
+                  "FunctionName" : module.lambda_function-treasuryReportGeneration["1C"].lambda_function_arn
```
TylerHendrickson, line 78: (see above)

terraform/treasury_generation_step_function.tf
```diff
@@ -0,0 +1,137 @@
...
+            }
+          }
+        ]
+      }
+    }
+  })
+
+  service_integrations = {
+    lambda = {
+      lambda = [module.lambda_function-treasuryReportGeneration["1A"].lambda_function_arn, module.lambda_function-treasuryReportGeneration["1B"].lambda_function_arn, module.lambda_function-treasuryReportGeneration["1C"].lambda_function_arn, module.lambda_function-subrecipientTreasuryReportGen.lambda_function_arn]
```
TylerHendrickson, line 132: (see above)

```suggestion
      lambda = [
        module.lambda_function-treasuryReportGeneration.lambda_function_arn, 
        module.lambda_function-subrecipientTreasuryReportGen.lambda_function_arn,
      ]
```

TylerHendrickson (changes requested):

terraform/treasury_generation_lambda_functions.tf
as1729: Keeping this lambda the same makes sense to me. The advantage I saw was mainly for observability in datadog. By filtering by function ARN it is easy to know whether the Project 1A/1B/1C files are failing to generate as opposed to adding a secondary filter on payload.

Since we can always add additional arguments to the log-context I'm good with making this change and simplifying this setup to have a single lambda.

as1729: I can see the need for different capacity requirements. I assume there are more of one type of project to generate over time, but I don't have any data at the moment to base this off-of so we can always separate it out when we notice this in the future.

TylerHendrickson: > I can see the need for different capacity requirements. I assume there are more of one type of project to generate over time, but I don't have any data at the moment to base this off-of so we can always separate it out when we notice this in the future.

FWIW, I think it will be a while before we're scaled to the point where we'll need to tune concurrency for these Lambda functions, the Lambda + SFN services should be able to work out scheduling executions on their own to a relatively high degree of scale.

TylerHendrickson (approved):
@as1729 One follow-up comment for you to consider; I disabled auto-merge for this PR in case you want to make any more changes, but if my suggestion doesn't apply, consider this approved to merge.

python/src/functions/generate_treasury_report.py
```diff
@@ -76,7 +75,7 @@ def handle(event: S3Event, context: Context):
 
     s3_client: S3Client = boto3.client("s3")
 
-    project_code = os.getenv("PROJECT_USE_CODE")
+    project_code = event["ProjectType"]
```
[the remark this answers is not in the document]
as1729: Makes sense, I'll update to remove the entire context in a separate PR since there's a few more changes I'm seeing that may be needed to that code.

PR #374, opened 2024-08-01
chore: update the certificate file to connect to RDS

as1729:


TylerHendrickson (changes requested):
@as1729 I left some testing notes for you to consider about the proper path to provide for the `sslcert` value in Lambda environments' connection strings.

**However,** even with the base path updated to `/var/task`, I think this won't work just yet, as `yarn rw build api` doesn't actually include any of the files under `api/db/` in the resulting `api/dist` contents, which means they aren't ultimately shipped in our Zip artifact. You can see for yourself by running the following commands ([which we use](https://github.com/usdigitalresponse/cpf-reporter/blob/7c4894ab902177ecdc8467958654efebbb3ab444/.github/workflows/build.yml#L444-L449) in our CI/CD pipeline's `Build API Lambda function zip artifacts` job) locally:
```cli
yarn rw build api
yarn zip-it-and-ship-it ./api/dist/functions ./api/dist/zipballs
```

If you inspect the resulting `api/dist/zipballs/graphql.zip` file (e.g. `zipinfo ./api/dist/zipballs | grep rds-global-bundle`), you'll see that the `.pem` file is mising.

I additionally verified that there are no `.pem` files in the Lambda zip artifact that was uploaded to S3 during the most recent Staging deployment.

I think there are three options:
1. Leave out the `sslmode=` parameter entirely and see if Prisma is able to find the correct CA bundle stored on the Lambda runtime environment's filesystem at `/etc/ssl/cert.pem`.
2. Provide `sslmode=/etc/ssl/cert.pem` explicitly – this file is included in the Lambda Node.js 20.x runtime environment, and is expected to provide RDS certificates.
3. Modify our `Build API Lambda function zip artifacts` job in the CI/CD pipeline by adding a command to manually add CA bundle file to the root directory of the zip files' contents as follows:
    ```yaml
          - name: Build the api side
            run: yarn rw build api
            env:
              CI: 1
          - name: Zip the built functions
            run: yarn zip-it-and-ship-it $BUILD_OUTPUT_PATH/functions $ARTIFACTS_PATH
          - name: Add RDS CA bundle to zipped Lambda artifacts
            run: |
              for $zipfile in $ARTIFACTS_PATH/*.zip; do
                zip -ju $zipfile api/db/rds-global-bundle.pem
              done;
    ```
    Once the zipfile is unpacked into a Lambda function's runtime environment, the file should be available at `/var/task/rds-global-bundle.pem`.

terraform/functions.tf
```diff
@@ -308,7 +308,7 @@ module "lambda_function-graphql" {
         "s3:HeadObject",
       ]
       resources = [
-        # Path: treasuryreports/{organization_id}/{reporting_period_id}/{filename}.csv
+        # Path: treasuryreports{organization_id}/{reporting_period_id}/{filename}.csv
```
TylerHendrickson, line 311: We discussed already but for posterity: looks like this change was unintentional.
```suggestion
        # Path: treasuryreports/{organization_id}/{reporting_period_id}/{filename}.csv
```

TylerHendrickson (changes requested):

TylerHendrickson (approved):

PR #384, opened 2024-08-14
chore: make tweaks to ensure step-function kick-off is consistent across file-generation lambdas

as1729:
- Add reporting data bucket as environment variables to the python lambdas
- Update project-lambda from S3 event structure to custom Payload structure
- Update both Subrecipient and Project Lambda to have a schema/type-defined payload structure for easier structure creation
- Add ability to upload Output template files from the graphql lambda
- Ensure Graphql and python lambdas have put/get object permissions for files within treasuryreports/output-templates and treasury reports… namespace of s3 bucket.
- Update tests based on the above changes for all the lambdas.

### DIFF
--- a/python/src/lib/constants.py
+++ b/python/src/lib/constants.py
@@ -1,0 +1,21 @@
+from enum import Enum
+
+from src.schemas.project_types import ProjectType
+
+
+class OutputTemplateFilename(Enum):
+    CPF1ABroadbandInfrastructureTemplate = "CPF1ABroadbandInfrastructureTemplate"
+    CPF1BDigitalConnectivityTechTemplate = "CPF1BDigitalConnectivityTechTemplate"
+    CPF1CMultiPurposeCommunityTemplate = "CPF1CMultiPurposeCommunityTemplate"
+    CPFSubrecipientTemplate = "CPFSubrecipientTemplate"
+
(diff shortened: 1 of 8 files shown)


python/src/lib/constants.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __str__(self):
+        return self.value
+
+
+OUTPUT_TEMPLATE_FILENAME_BY_PROJECT = {
+    ProjectType._1A: OutputTemplateFilename.CPF1ABroadbandInfrastructureTemplate,
+    ProjectType._1B: OutputTemplateFilename.CPF1BDigitalConnectivityTechTemplate,
+    ProjectType._1C: OutputTemplateFilename.CPF1CMultiPurposeCommunityTemplate,
+    "Subrecipient": OutputTemplateFilename.CPFSubrecipientTemplate,
```
vshia, line 20: If this is used alot, might be helpful to move this into a constant value.

vshia (approved):
Just one minor comment but this makes alot of sense to me.

python/src/lib/constants.py
as1729: Good call, I'll create a separate PR for this shortly

PR #388, opened 2024-08-16
Issue #318 - Lambda Function To Create Archives

tzinckgraf:
This is a fix for issue #318 . It is a lambda function that zips a bunch of treasury files into one zipped file and pushes that to S3.

The bulk of this PR is functions and scripts to test step functions locally.

This logic for the archive function is based on logic from @jakekreider . He was the original author of this work.


python/src/functions/create_archive.py
```diff
@@ -0,0 +1,91 @@
...
+    organization_ids = {p["organization_id"] for p in payloads if p["organization_id"]}
+    reporting_period_ids = {p["reporting_period_id"] for p in payloads if p["reporting_period_id"]}
+
+    if len(organization_ids) != 1:
+        raise ValueError("All payloads must have the same organization_id")
+    if len(reporting_period_ids) != 1:
+        raise ValueError("All payloads must have the same reporting_period_id")
+
+    organization_id = organization_ids.pop()
+    reporting_period_id = reporting_period_ids.pop()
```
as1729, line 38: Small change requested- can we update this to follow the same pattern of the other lambda functions where the payload gets validated using `pydantic` which will avoid having to manually generate these errors?

Suggested code may be like
```py
from src.lib.treasury_generation_common import OrganizationObj
from pydantic import BaseModel

class CreateArchivePayload(BaseModel):
    organization: OrganizationObj

try:
    payload = CreateArchivePayload.model_validate(event)
except...

create_archive(payload.organization.id, payload.organization.preferences.current_reporting_period_id, boto3.client("s3"), logger)
```

Example: https://github.com/usdigitalresponse/cpf-reporter/blob/main/python/src/functions/subrecipient_treasury_report_gen.py#L59-L63

python/src/functions/create_archive.py
```diff
@@ -0,0 +1,91 @@
...
+import tempfile
+from typing import Any
+import zipfile
+
+import boto3
+from aws_lambda_typing.context import Context
+from mypy_boto3_s3.client import S3Client
+from src.lib.logging import get_logger, reset_contextvars
+
+S3_BUCKET = os.environ.get("S3_BUCKET", "test_bucket")
```
as1729, line 11: confirming that the name of the bucket here will be as `os.environ.get("REPORTING_DATA_BUCKET_NAME", "test_bucket")

python/src/functions/create_archive.py
```diff
@@ -0,0 +1,91 @@
...
+        "statusCode": 200,
+        "Payload": {
+            "organization_id": organization_id,
+            "reporting_period_id": reporting_period_id,
+        },
+    }
+
+
+def create_archive(
+    org_id: str, reportiong_period_id: str, s3_client: S3Client, logger=None
```
as1729, line 53: typo `reportiong_period_id` -> `reporting_period_id`

localstack/entrypoint/init-aws.sh
```diff
@@ -13,56 +13,3 @@ for email in "${VALID_EMAILS[@]}"; do
 done
 
 awslocal s3api create-bucket --bucket cpf-reporter --region us-west-2 --create-bucket-configuration '{"LocationConstraint": "us-west-2"}'
-
```
as1729, line 16: is this deletion intended?

as1729 (changes requested):

as1729:
Thanks @tzinckgraf ! I added some suggested changes based on some things we changed in the other lambda functions over the past week.

python/src/functions/create_archive.py
tzinckgraf: Done.

python/src/functions/create_archive.py
tzinckgraf: Makes sense. Code is updated to use pydantic models with test cases.

localstack/entrypoint/init-aws.sh
tzinckgraf: It is. I tend to removing trailing whitespace from the end of files. I can add it back in if you prefer.

as1729: apologies I didn't intend to select this specific line was referring to lines 16-68. Will create a separate thread.

localstack/entrypoint/init-aws.sh
```diff
@@ -13,56 +13,3 @@ for email in "${VALID_EMAILS[@]}"; do
...
-                    "input.$": "$"
-                }
-            },
-            "OutputPath": "$.Payload",
-            "End": true
-        }
-    }
-  }' \
-  --role-arn "arn:aws:iam::000000000000:role/stepfunctions-role"
-
```
as1729, line 68: @tzinckgraf this is the deletion I was referring to. This particular script was helpful for me to test things in local stack but if you think we can do this differently I'm good with removing it.

tzinckgraf: I ended up moving most of that to the script `python/tests/test-create-archive.sh`. I found it was easier to rerun and tweak in that script, and it was a bit hidden in the init-aws script. I can move it back if you prefer.

as1729: ah gotcha no worries! I missed this update makes sense 👍

as1729 (approved):

PR #403, opened 2024-08-28
chore: refactor common code across lambda functions and fixes incorrect typing

as1729:
This PR was created to fix two things:

1. Typing errors within the python code related to `TemporaryFileWrapper`.
2. Issue with the usage of `json.dump`. Unclear why this wasn't generating test failures but the function call was raising a type-issue for a missing argument.
3. Consolidates the generation for s3 file keys for output files and stores it within `treasury_generation_common.py`. While this doesn't reduce the lines of code- I'm hoping that consolidating this reduces the chance of having a malformed object key anywhere else in the code as it appears to be recreated in multiple places.

### DIFF
--- a/python/src/functions/generate_treasury_report.py
+++ b/python/src/functions/generate_treasury_report.py
@@ -12,12 +12,13 @@ from openpyxl import Workbook, load_workbook
 from openpyxl.worksheet.worksheet import Worksheet
 from pydantic import BaseModel
 
-from src.lib.constants import OUTPUT_TEMPLATE_FILENAME_BY_PROJECT
 from src.lib.logging import get_logger, reset_contextvars
 from src.lib.s3_helper import download_s3_object, upload_generated_file_to_s3
 from src.lib.treasury_generation_common import (
     OrganizationObj,
+    OutputFileType,
     UserObj,
+    get_generated_output_file_key,
     get_output_template,
 )
 from src.lib.workbook_utils import convert_xlsx_to_csv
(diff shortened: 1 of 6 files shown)


python/src/functions/generate_treasury_report.py
```diff
@@ -215,16 +223,24 @@ def process_event(payload: ProjectLambdaPayload, logger):
...
+                key=get_generated_output_file_key(
+                    file_type=OutputFileType.CSV,
+                    project=project_use_code,
+                    organization=organization,
+                ),
                 file=csv_file,
             )
         # Store project_id_agency_id to row number in a json file
         with tempfile.NamedTemporaryFile("w") as json_file:
-            json_file = json.dump(project_agency_id_to_row_map)
```
as1729, line 223: @vshia would be great to get your input here. I updated this based on typing errors I noticed. However I wasn't sure if this breaks any of the existing functionality.

python/src/functions/generate_treasury_report.py
```diff
@@ -177,6 +178,10 @@ def process_event(payload: ProjectLambdaPayload, logger):
             with tempfile.NamedTemporaryFile() as file:
                 # Download projects from S3
                 created_at = file_info.createdAt
+
+                # TODO: Add a verification to ensure that the objectKey here is formatted correctly.
+                # Path must be: uploads/{organization_id}/{agency_id}/{reporting_period_id}/{upload_id}/{filename}.xlsm
```
laurelkorwin, line 183: You could potentially borrow logic here from `matchRegex` in `processValidationJson`, which has the same key structure

python/src/lib/treasury_generation_common.py
```diff
@@ -22,6 +23,17 @@ class UserObj(BaseModel):
...
     email: str
 
+class OutputFileType(Enum):
+    XLSX = "xlsx"
+    CSV = "csv"
+    JSON = "json"
+
+
+def get_generated_output_file_key(
+    file_type: OutputFileType, project: str, organization: OrganizationObj
```
laurelkorwin, line 33: any appetite for making `project` values an Enum too? I know we're grabbing from the dictionary below so perhaps not strictly necessary but could make the flow here more legible?

laurelkorwin (approved):

python/src/lib/treasury_generation_common.py
as1729: 💯  - I'll create a separate PR with this updated

python/src/functions/generate_treasury_report.py
as1729: 💯 - will address in a separate PR

PR #407, opened 2024-09-03
Jest testing env setting

greg-adams:
- This PR explicitly sets `NODE_ENV` to `test` to avoid any conflict with the Docker container while running unit tests in development.  Jest would normally set this automatically, however it will not update it if it has been set. Now that we are referencing this variable in code, some tests fail locally due to inheriting the value of `development` as set in docker compose file.


greg-adams:
@as1729 opening this pr to hold if any discussion is needed on this approach

as1729:
@greg-adams when running the redwood test suite here is what I'm doing:

1. Run this command to ssh into the `redwood` container: `docker compose -f ./docker-compose.dev.yml run --rm -it redwood /bin/bash`
2. Run yarn command to kick-off tests: `yarn rw test`

On my machine running `yarn rw test` automatically uses the test environment and ensures that the suite is run on the `TEST_DATABASE_URL` etc.

Is the issue you are experiencing that when you run `yarn rw test` the `redwood` container is _not_ using the test-database?

greg-adams:
@as1729 I don't appear to have issues with the db, just the value of `process.env.NODE_ENV` being set as 'development' when running tests in the container. It should be 'test', and a value of 'development' causes some tests to fail since we now use this variable in the code

as1729 (approved):

as1729:
Got it @greg-adams I think I misunderstood what was happening here and was concerned that the `NODE_ENV` would be `test` for all processes being run within the container and not just the tests.

PR #414, opened 2024-09-05
fix: Fix breaking change in actions/upload-artifacts 4.4.0

as1729:
Same as https://github.com/usdigitalresponse/usdr-gost/pull/3469


TylerHendrickson (changes requested):
@as1729 There are a 3 instances of `actions/upload-artifacts` in `build.yml` and 1 instance in `terraform-plan.yml` that this PR still needs to address.

TylerHendrickson (approved):

PR #416, opened 2024-09-06
fix: ensure object-path is defined as a condition and not on the resource-name

as1729:
We are continuing to see 403 errors on the treasury report generation similar to [this](https://app.datadoghq.com/logs?query=service%3Acpf-reporter&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAZHEsFz7yNzgHAAAAAAAAAAYAAAAAEFaSEVzRjdUQUFCbjFXUC1WUFFvLVFBQwAAACQAAAAAMDE5MWM0YjAtNzdlNi00NjZmLWIyNDQtYmNiM2JhMjY1MzA0&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1725578260233&to_ts=1725581860233&live=true)

The reason for this is that the `ListBucket` operation must be defined on a resource-name rather than a resource-name plus path. In order to define the specific path, I need to add it as a condition.

I was able to test this working in my personal AWS account so feel much more confident about this change resolving these issues.

### DIFF
--- a/terraform/treasury_generation_lambda_functions.tf
+++ b/terraform/treasury_generation_lambda_functions.tf
@@ -29,11 +29,11 @@ module "lambda_function-subrecipientTreasuryReportGen" {
       resources = [
         "${module.reporting_data_bucket.bucket_arn}",
       ]
+      # Path: /{organization_id}/{reporting_period_id}/*
       condition = {
-        StringEquals = {
-          # Path: /{organization_id}/{reporting_period_id}/*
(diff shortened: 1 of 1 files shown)


TylerHendrickson (approved):
Needs `terraform fmt` but the substance looks good.

vshia (approved):

TylerHendrickson (approved):

PR #477, opened 2024-10-04
Speedups and enhancements for GitHub Actions CI workflows

TylerHendrickson:
This PR addresses a few issues currently affecting continuous integration workflows for this repo:

- **`step-security/harden-runner` actions taking too long.** Fixed by upgrading all usages of this action to the latest version.
- (Side note: it's worth mentioning that the gains from "Prepare for QA" would be more substantive if we enabled Dependabot, since 
- **"Prepare for QA" step doesn't provide an effective Node cache.** This was happening because although `actions/setup-node` persists the _Yarn_ cache between steps, it does not include `node_modules` – presumably because reusing `node_modules` across different versions of Node can cause problems. A separate usage of `actions/cache` has been added to implement the following caching behaviors:
  - The cache key is created by hashing the contents of `yarn.lock`.
  - When "Prepare for QA" starts, it attempts to restore a `node_modules` cache created from the current state of `yarn.lock`. If this results in a cache miss, then a new cache is created after installing dependencies. When a cache hit is possible, `yarn install` only takes about 10 seconds to complete. This behavior ensures that a `node_modules` cache always exists after "Prepare for QA" completes. 
    - **Important note:** We should bear in mind that this could cause problems if `yarn install` creates a corrupt `node_modules` for some reason, which would normally be fixed on a subsequent run of `yarn install`. While this shouldn't happen, it might(?). In these cases, the cache would need to be manually deleted from GitHub. A safer alternative would be to never restore cache in "Prepare for QA" and instead always build a fresh one, although this would slow things down (similar to the status quo).
  - Each dependency of "Prepare for QA" restores the `node_modules` cache and **does not run `yarn install`**.
  - Altogether, these changes shave about 2 minutes of runtime from the Continuous Integration workflow. Given the consideration I noted above regarding cache spoilage, it might be justifiable to re-add a `yarn install` step to these dependent jobs, which would run _after_ cache restoration. In most cases, it would merely add ~10 seconds to each job (basically how much time it takes for Yarn to decide not to do anything), but the benefit would be that it would give Yarn a chance to fix things in case the `node_modules` cache was spoiled.
    - @as1729 curious for your thoughts here.
- **Test coverage reports get cut off in PR comments.**
  - "Terraform Summary" and "QA Summary" PR comments now provide a direct link to the workflow at the top, since they currently disappear when a comment is truncated due to length.
  - Each test suite's coverage report is posted to a different step summary, in addition to the "QA Summary" comments left on PRs. This ensures that all 3 coverage reports can still be viewed, even if the "QA Summary" comment is truncated.
  - `*.sdl.ts` files are excluded from Jest coverage collection, since they are not tested and otherwise taking up space in the coverage reports.

Altogether, these changes reduce Continuous Integration workflow execution by about 8-10 minutes (and up to 12 minutes, according to some hasty napkin math). Most of this is due to upgrading the `step-security/harden-runner` action, which is a good argument for enabling automatic dependency upgrades for this repo 🤓 

PS - You won't see any of these gains on this PR, because we only run GitHub actions using the base branch configuration. For a preview of the effective changes, see #475.


TylerHendrickson:
@as1729 I noticed a few times in #475 that the API-side test suite was a little flaky – here's [one example](https://github.com/usdigitalresponse/cpf-reporter/actions/runs/11172619239/job/31059384586). Is this a known issue, or should I look into whether that's unique to the changes introduced by this branch?

as1729:
> Is this a known issue, or should I look into whether that's unique to the changes introduced by this branch?
This is a known issue captured in #476 and @greg-adams is working to resolve it.

as1729:
> Given the consideration I noted above regarding cache spoilage, it might be justifiable to re-add a yarn install step to these dependent jobs, which would run after cache restoration. In most cases, it would merely add ~10 seconds to each job (basically how much time it takes for Yarn to decide not to do anything), but the benefit would be that it would give Yarn a chance to fix things in case the node_modules cache was spoiled.

If its a 10second time-add and gives us higher confidence that we won't be dealing with corrupt packages it seems reasonable .However, I'm not able to think of any reasons why these packages might get corrupted from one step to the next. If you think there are several ways this could happen that is out of our control, I'm good with adding a redundancy here.

> which is a good argument for enabling automatic dependency upgrades for this repo 🤓

💯  agreed. I thought @dependabot was already in this repository? example:  #439

as1729 (approved):

TylerHendrickson:
> > Given the consideration I noted above regarding cache spoilage, it might be justifiable to re-add a yarn install step to these dependent jobs, which would run after cache restoration. In most cases, it would merely add ~10 seconds to each job (basically how much time it takes for Yarn to decide not to do anything), but the benefit would be that it would give Yarn a chance to fix things in case the node_modules cache was spoiled.
> 
> If its a 10second time-add and gives us higher confidence that we won't be dealing with corrupt packages it seems reasonable .However, I'm not able to think of any reasons why these packages might get corrupted from one step to the next. If you think there are several ways this could happen that is out of our control, I'm good with adding a redundancy here.

My thinking is that if a package is broken on NPM and then updated without a version bump (rare, but it happens), our cache would be stuck on the broken package. Maybe network availability is another risk, although I'm not sure if Yarn would complete successfully if that were the case. 

Maybe we leave as-is for now and keep this as a back-pocket strategy if we lose confidence in passing around `node_modules`.

> > which is a good argument for enabling automatic dependency upgrades for this repo 🤓
> 
> 💯 agreed. I thought @dependabot was already in this repository? example: #439

We have Dependabot enabled for security updates (when a known vulnerability has a compatible patch) but not for general dependency updates.

as1729 (approved):

PR #490, opened 2024-10-11
fix: ensure treasury generation payload has information for all projects

as1729:
#489


as1729:
Merging this PR in order to unblock partners in their report generation. Will address any comments in a separate PR.

PR #586, opened 2024-12-11
[treasury-report] treasury archive timestamps in the emails

nowei:
Adds the timestamp of when the zip archive was last generated into the send treasury report emails.


nowei:
@as1729 I'm not sure if this was what the ticket was asking for? 

I think it might technically be more accurate to use the last modified timestamps of the files that were used to generate the treasury report zip, but I'm not sure how to walk the file structure to get that information. 

In any case, the time that the zip is generated should generally be after the files were updated, so it still provides a signal as to the freshness of the file?

as1729 (approved):

